Issue #36, opened 2020-05-05
Setup a ci pipline - probably github actions


brafdlog:
CI should currently run tests.

In the future can also package the application for installation

PR #49, opened 2020-06-02
Create Github Action for PullRequest validation

baruchiro:
related to #36

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,25 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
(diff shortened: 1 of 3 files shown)


baruchiro:
For the tests, we need first to merge #48, and take care of the `config` file in `EmailSender`.

Now we expecting to `config` file, but it failing in the build time because it does not exist. We can use environment variables instead, or add the `config` file with placeholder values, and then tell Git to stop track on it (I did it in my repo, but it can be a confusing solution.)

baruchiro:
Irrelevant after merging israeli-bank-scrapers-desktop repo (#50)

PR #53, opened 2020-06-07
Try to make GithubActions pass

baruchiro:


brafdlog:
This is a small pr and will fix the ci, so do want to merge it

package.json
```diff
@@ -11,11 +11,12 @@
     "build": "vue-cli-service electron:build",
     "postinstall": "electron-builder install-app-deps",
     "postuninstall": "electron-builder install-app-deps",
-    "lint": "eslint --ext .js,.vue -f ./node_modules/eslint-friendly-formatter src",
-    "lint:fix": "eslint --ext .js,.vue -f ./node_modules/eslint-friendly-formatter --fix src",
+    "prepare": "ncp src/originalBudgetTrackingApp/categoryCalculationScript-example.js src/originalBudgetTrackingApp/categoryCalculationScript.js",
```
brafdlog, line 14: I think ncp is an overkill here, cp -n will do the trick

package.json
```diff
@@ -12,11 +12,11 @@
     "postinstall": "electron-builder install-app-deps",
     "postuninstall": "electron-builder install-app-deps",
     "prepare": "cp -n src/originalBudgetTrackingApp/categoryCalculationScript-example.js src/originalBudgetTrackingApp/categoryCalculationScript.js || true",
-    "lint": "yarn prepare && eslint --ext .js,.vue -f ./node_modules/eslint-friendly-formatter src",
```
brafdlog: Prepare runs automatically after install so no need to invoke it explicitly

package.json
baruchiro: `cp -n` shouldn't work on Windows, and also the `|| true` syntax, I don't know how it works on *Github Actions*.

On my desktop, it didn't work.

brafdlog: ah, forgot about windows... Ok, will change that

PR #54, opened 2020-06-10
Use config manager in Importers

baruchiro:
In this change, I'm using **Vuex** (the official state manage for Vue) to hold and represent the `config` object from `configManager`.

Then, I'm using `vuex-persist`. This is a little library to create a **Vuex Plugin**- a function that called every time the state is changed. With `vuex-persist` I can define a handler for a specific module, so I created a handler for the `config` Vuex module, to load this state from `configManager` and save it with `configManager` every time it changes.

This leads us to the fact that the **state** will always be the same as the **config**. We have to remember this in both directions.


brafdlog:
The consistent naming convetion in the repo is camelCase for dirs and files so I would name the dir configManager

src/modules/scrapers.js
```diff
@@ -1,14 +1,7 @@
 /* eslint-disable max-len */
-import { createScraper, SCRAPERS } from 'israeli-bank-scrapers-core';
+import { createScraper } from 'israeli-bank-scrapers-core';
 import getChrome from './downloadChromium';
```
brafdlog, line 4: This change broke the import in Launch.spec.js

src/components/MainPage/Importers/Importer.vue
```diff
@@ -76,34 +68,23 @@ export default {
   data() {
     return {
       importing: false,
+      // TODO: show browser should be in config. Also maybe one for all accounts
```
brafdlog, line 71: Yes and yes. In general I think the common usecase is scrape all importers and then this type of configuration is global

test/e2e/specs/Launch.spec.js
```diff
@@ -1,15 +1,16 @@
...
 
 const screenshotsDir = './screenshots';
 
-jest.setTimeout(1000000);
+jest.setTimeout(100000);
 
 // Remove when https://github.com/nklayman/vue-cli-plugin-electron-builder/issues/625 closed
-const skip = process.env.GITHUB_ACTIONS && process.platform === 'win32';
+// const skip = process.env.GITHUB_ACTIONS && process.platform === 'win32';
+const skip = true;
```
brafdlog, line 13: The test got stuck both locally and in the ci. When debugging locally it seems it got stuck  running `testWithSpectron`. Didn't even get to the tests themselves.

Issue #51, opened 2020-06-05
Merge with the israeli-bank-scrapers-desktop repo


brafdlog:
Project for issues we want to handle pre merge: https://github.com/brafdlog/budget-tracking/projects/4

Related issues:
- [x] Make the configManage use the crypto module and remove hard coded encryption key
- [x] Make the node code expose an api as defined [here](https://miro.com/app/board/iXjVOf3EJUs=)
- [x] Update the electron UI flow to work with two main flows - configure, run scraping as defined [here](https://miro.com/app/board/iXjVOf3EJUs=)
- [x] Make the electron UI work the node "library" code
- [x] Remove the Contributors part or hide it in a Menu or something #87 
- [x] Align filename #57

Post Merge Actions:
 - [ ] Decide how to release (by `Tag` or by pushing to `master`)
 - [ ] Copy relevant issues from `israeli-bank-scrapers-desktop`.
 - [ ] New repo/app name?
 - [ ] Check if there unused packages
 - [ ] Add MacOS to tests
 - [ ] Create Splash screen (until the renderer process loaded)
 - [ ] Set show browser in `configManager`? Define it for all importers?
 - [ ] Add ability to disable importer and exporter (and keep them configured)

baruchiro:
Note that the folder `nuxt` contains the [Github Pages](https://baruchiro.github.io/israeli-bank-scrapers-desktop/), do you want to keep it?

I think the regular users need a beautiful webpage in Hebrew, with a big download button and help content.

But if you want to rethink that, maybe to use another technology...

baruchiro:
About the file structure, I already started to reorg the project in https://github.com/baruchiro/israeli-bank-scrapers-desktop/pull/247, but it is not important, since I think the most of my code will move to BudgetTrackingApp, or just gone, and the rest of the code is the `components` folder.

brafdlog:
Regarding the githup pages, you mean you wanted to use something else instead?

I am fine with removing it.

brafdlog:
> About the file structure, I already started to reorg the project in [baruchiro/israeli-bank-scrapers-desktop#247](https://github.com/baruchiro/israeli-bank-scrapers-desktop/pull/247), but it is not important, since I think the most of my code will move to BudgetTrackingApp, or just gone, and the rest of the code is the `components` folder.

Ok, I updated the todo to be to remove the code that is no longer relevant when the ui starts using the budget tracking code.

baruchiro:
[...]

PR #59, opened 2020-07-14
Output vendor interface

baruchiro:
I started to define an interface for the output vendors.

I created a `Json` exporter, and now it is shown in the UI, but it still doesn't work- the next step is to remove the `MainTable` and replace it with a big **run** button.

So I think we can merge this, and continue to work on

 - Implement the interface for the rest of the output vendors (it will require more decisions about the interface)
 - See all the TODOs I wrote, maybe before merging this PR.
 - Create a big **run** button.
 - Implement the full process- from importing to exporting.

I will start with the TODOs and then with the **run**.
Maybe you @brafdlog can convert the output vendors to the new interface, because you wrote them.


src/App.vue
```diff
@@ -3,6 +3,7 @@
 </template>
 
 <script>
+// TODO why we need this component?
```
baruchiro, line 6: We creating the root component in `main.js`. The root component used the `App.vue` component, and this component do nothing except the same thing- use the next component.

So I'm removing this.

baruchiro: Actually we doing so many changes, no need to describe each one.

src/store/index.js
```diff
@@ -4,12 +4,14 @@ import migrations from './migrations';
...
 
-
 Vue.use(Vuex);
 
 const store = new Vuex.Store({
   modules,
-  plugins: [configPersistPlugin('config')],
+  // TODO: this plugin just copy the config state,
+  // if the backup will fail, we will have two configs- filesystem and store
+  // Maybe we need to backup in the action
```
baruchiro, line 13: See https://github.com/brafdlog/budget-tracking/issues/51#issuecomment-658835043
I'm keeping this comment until we discuss and decide if we want to change this.

src/components/shared/FormField/index.js
```diff
@@ -0,0 +1,14 @@
...
+
+const typeToComponent = {
+  text: textField,
+  file: fileField,
+};
+
+export default {
+  functional: true,
+  // TODO do we want to inject all props from 'outputVendors' (by passing context.data),
+  // or we want to sanitize and select specific properties?
```
baruchiro, line 12: The `context.data` is the object from `v-bind`. In my code, I injected the whole `fields` data from `outputVendor`.

The question is if we want to sanitize or filter some properties from the `fields` before it affects the UI.
Because now, I'm giving to `outputVendor` developer full control on the UI, because what he will define in the `fields` will become to component properties.

src/main.js
```diff
@@ -1,30 +1,50 @@
...
 import LoggerPlugin from './plugins/logger';
 import vuetify from './plugins/vuetify';
 import router from './router';
+import Splash from './Splash';
 import store from './store';
 
 initializeReporter();
 
+process.on('unhandledRejection', (error) => {
+  debugger;
```
brafdlog, line 16: Probably don't want a debugger here :)

baruchiro, line 16:
```suggestion
```

src/store/index.js
```diff
@@ -4,12 +4,15 @@ import migrations from './migrations';
 import modules from './modules';
 import configPersistPlugin from './plugins/persisted-config';
 
-
 Vue.use(Vuex);
 
 const store = new Vuex.Store({
   modules,
-  plugins: [configPersistPlugin('config')],
+  // TODO: https://github.com/brafdlog/budget-tracking/issues/51#issuecomment-658835043
```
brafdlog, line 11: See comment in this thread: https://github.com/brafdlog/budget-tracking/issues/51#issuecomment-658835043

src/originalBudgetTrackingApp/outputVendors/index.js
```diff
@@ -0,0 +1,10 @@
+import path from 'path';
+
+const files = require.context('.', true, /\.\/.+\/index\.js$/);
```
brafdlog, line 3: I prefer a less dynamic approach. Especially in apis. We are not going to be adding new vendors every day so the value of dynamic code is low IMO. I will spend some time on the output vendors code

src/store/modules/Config.js
```diff
@@ -1,39 +1,28 @@
...
+  addImporter: (state, importer) => state.scraping.accountsToScrape.push(importer),
+  removeImporter: (state, importerId) => {
     state.scraping.accountsToScrape = state.scraping.accountsToScrape.filter((importer) => importer.id !== importerId);
   },
+  addExporter: (state, { name, ...values }) => {
+    state.outputVendors[name] = values;
+  },
 };
 
 // TODO not sure we need that. We need a status but not in the state
```
brafdlog, line 22:
```suggestion
```

src/components/shared/FormField/index.js
brafdlog: As I wrote in another comment, I think we should less dynamic in the outputVendors. There won't be that many. I think they should have a clear interface for their common functionality and extentions for vendor specific behaviour and the ui will match this - common component for the common code and specific components for vendor specific behaviour.
That is what I think at the moment, I will spend some time on the output vendor api and then will be smarter

brafdlog:
@baruchiro As we discussed this pr is now in a state where the application is working and ts is set correctly. I suggest we merge it and start a new branch for the output vendors and types. WDYT?

baruchiro:
Right, let's merge it

PR #60, opened 2020-07-16
Update README.md

antmir:
adding a mandatory disabling of integration not used to the readme


brafdlog:
Thanks for the contribution @antmir 😄

PR #62, opened 2020-07-16
Update bankScraper.js

antmir:
don't throw error on credentials that don't have username (like isracard that use id and not username)


brafdlog:
Good point and thanks for the contribution.

Since this #50 includes major changes I suggest creating future prs from the the unifyRepos branch until we merge it

PR #63, opened 2020-07-16
Add typescript

brafdlog:
This is a work in progress.

Added typescript config for the "library" code. The current code contains the configuration and change of files to ts without defining any types yet. Once we are happy with this I will define the types, remove the ts ignores and make the ts config more strict.

Typescript is supposed to be supported pretty much out of the box in vue but at the time when I tried to set it up for the whole project I ran into problems, so i enabled it for the library code only. I probably missed something small and we can enable it for the whole project.


baruchiro:
We will need to discuss how much strict we want.
We can use TS as hinting, means that not fail almost anything, but add types only when we want, or use strict config to force declare typing everywhere.

Of course, for migration, the strict level will be lower.

And I will agree with whatever you decide. But for my opinion, I think the lower level is the best because you can use types only when they help.

baruchiro:
I updated some `@ts-ignore` cases in my branch, see the diff here https://github.com/brafdlog/budget-tracking/compare/addTypescript2...addTypescript2-baruch2 (the diff includes a merge from `unifyRepos`). You can merge them with you this branch or if you prefer I can open a new PR after this will merge.

I see you didn't convert the `modules` library to TS. I tried to convert it, but since you set the `strict` to be `true`, there are a lot of problems, so maybe we can keep the `modules` library in `js` until it will be deprecated.

brafdlog:
I merged your pr into this one.

Regarding the ts-ignores, I want to have the typsecript setup and then will work on all the types in the #59 pr.

Regarding strictness of ts, I prefer to have a strict configuration. I think ts is most valuable when you can rely on the types. In the cases where it is more hassle than value we can add ts ignore but on the regular flow I think we should default to the strict flow.

Let's discuss this after I define the main types for the library code and reach a decision.

PR #66, opened 2020-07-24
Make eslint work with typescript and small fixes

brafdlog:


src/main.ts
```diff
@@ -17,7 +17,6 @@ process.on('unhandledRejection', (error) => {
   console.error(error);
```
baruchiro, line 17: > Unexpected console statement

brafdlog: Removed in the addTypes pr that I will open in a sec

PR #67, opened 2020-07-28
Add types to originalBudgetTrackingApp code

brafdlog:


package.json
```diff
@@ -24,13 +26,13 @@
...
   },
   "dependencies": {
+    "@brafdlog/israeli-bank-scrapers-core": "^0.9.2",
     "@getstation/electron-google-oauth2": "2.1.0",
     "@sentry/electron": "^1.3.0",
     "core-js": "^3.4.4",
     "download-chromium": "^2.2.0",
     "electron-log": "^4.1.1",
     "googleapis": "^49.0.0",
-    "israeli-bank-scrapers-core": "^0.9.1",
```
baruchiro, line 33: Are you planning to merge your package?

src/logger.ts
```diff
@@ -8,11 +8,11 @@ declare module 'electron-log' {
   }
 }
 
-export default function CreateLogger(app) {
+export default function CreateLogger(app: any) {
```
baruchiro, line 11: I think the `app` type is `Electron` or something (I'm from phone)

src/main.ts
```diff
@@ -13,10 +13,6 @@ import store from './store';
 
 initializeReporter();
 
-process.on('unhandledRejection', (error) => {
```
baruchiro, line 16: Why?
In this way we logging async exceptions

src/modules/downloadChromium.js
```diff
@@ -1,14 +1,19 @@
...
+
 export default async function downloadChromium(installPath, onProgress) {
-  onProgress({ percent: 0.02, message: 'Step 1: Downloading Chrome...' });
-  // onProgress: track download progress. receives one argument { percent, transferred, total }
-  return download({
-    revision,
-    installPath,
-    onProgress,
-  });
+  if (!chromePath) {
```
baruchiro, line 9: I understand that this is for download chrome just in the first time. But how has it worked so far? It also downloaded it only for the first time.

package.json
brafdlog: Until israeli-bank-scrapers  release the typescript branch. What other option do we have?

src/main.ts
brafdlog: I thought you asked to remove it in the other pr(#66). I think that if you don't have an unhandledRejection handler it will log the error anyway, but I will revert this change. So what did you mean in this comment https://github.com/brafdlog/budget-tracking/pull/66#discussion_r460475651?

src/modules/downloadChromium.js
brafdlog: Thats true, i see now that the library will prevent it from being downloaded twice. Reverting the change

package.json
```diff
@@ -14,7 +14,9 @@
     "prepare": "ncp src/originalBudgetTrackingApp/categoryCalculationScript-example.ts src/originalBudgetTrackingApp/categoryCalculationScript.ts || true",
     "lint": "eslint --ext .js,.vue,.ts -f ./node_modules/eslint-friendly-formatter src",
     "lint:fix": "yarn lint --fix",
+    "postlint": "yarn typeCheck",
```
baruchiro, line 17: I personally preferred to use explicit scripts instead of `pre`/`post` scripts (except `postinstall` and other special cases)

vue.config.js
```diff
@@ -9,6 +9,15 @@ const defineGlobals = (config) => {
     args[0] = defined;
     return args;
   });
+  config.module
+    .rule('babel')
+    .test(/base-scraper\.js$/)
```
baruchiro, line 14: What is it?

Did you forget to remove it?

`base-scraper` is from `israeli-bank-scrapers`

package.json
baruchiro: I don't know.... because I don't know exactly what you did.

I think in TypeScript you can declare untyped modules yourself, like we are doing in the `.d.ts` files.

https://www.typescriptlang.org/docs/handbook/modules.html#working-with-other-javascript-libraries

src/main.ts
baruchiro: I just notice that `lint` mark it as an error. I'm seeing this in the Github interface:

image

You just need to add ESLint ignore

package.json
brafdlog: I took the typescript branch in israeli bank scrapers, ran the build and published that as a package. So, when they merge the typescript brach we should just be able to replace my forked package to theirs.
We could start defining modules ourselves but since they already ha a typescript branch that seems to be close to merging i prefer to work over that code instead of doing double work. Makes sense?

baruchiro: Yes yes, you are absolutely right, we don't need to do more work here.

src/main.ts
brafdlog: ah, I didn't see this error because it is set to be ignored in dev environment.

I am changing this to be a warning in non prod environment so it will be more visible.

vue.config.js
brafdlog: Thats actually an interesting one. The built version of israeli bank scrapers in the typescript branch has a class with a property on it which fails the build so this was targeted at that, but I could add this to all js files.

brafdlog: But adding this to all js files makes the build much longer. I think a better solution is to update electron which will have a more up to date node version and won't need this. But I don't want to mix an electron update in this pr.

package.json
brafdlog: I don't have a preference on that. Changed

PR #71, opened 2020-08-07
Make the release run on mac as well

brafdlog:


baruchiro:
I think it's too early, but it does not matter.

If I remember correctly, there is a limit on the amount of MacOS machine usage, be aware of that.

brafdlog:
> I think it's too early, but it does not matter.
> 
> If I remember correctly, there is a limit on the amount of MacOS machine usage, be aware of that.

Really? I didn't see that in the docs. I only see a limitation on concurrent mac os jobs.
Since I work have a mac i would like to have a released version on a mac.

PR #72, opened 2020-08-16
Fix Scraping process

Arielgordon123:
image

### DIFF
--- a/src/originalBudgetTrackingApp/bankScraper.ts
+++ b/src/originalBudgetTrackingApp/bankScraper.ts
@@ -13,7 +13,7 @@ export const inputVendors = Object.keys(SCRAPERS)
 
 interface ScrapeParameters {
   companyId: AccountToScrapeConfig['key'];
-  credentials: AccountToScrapeConfig['credentials'];
+  credentials: AccountToScrapeConfig['loginFields'];
   startDate: Date;
   showBrowser?: boolean
 }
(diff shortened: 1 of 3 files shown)


src/originalBudgetTrackingApp/bankScraper.ts
```diff
@@ -13,7 +13,7 @@ export const inputVendors = Object.keys(SCRAPERS)
 
 interface ScrapeParameters {
   companyId: AccountToScrapeConfig['key'];
-  credentials: AccountToScrapeConfig['credentials'];
+  credentials: AccountToScrapeConfig['loginFields'];
```
baruchiro, line 16:
```suggestion
  credentials: AccountToScrapeConfig.loginFields;
```

Without looking at the code, I think if you access to the property itself, it will force the type-checking, right?
I think if we had done it from the beginning, the bug would have been discovered during compilation, no?

baruchiro:
Can you try to describe what the bug was?
Because I just see you changed the `credentials` to `loginFields`, and I don't know why to change the term?
My guess is the the `loginFields` comes from `israeli-bank-scrapers` and we transform it to `credentials`. Am I right?

Arielgordon123:
yes you are right 
for example, take a look here:
https://github.com/brafdlog/budget-tracking/blob/93b2512e9777621f037a2b9202daebe908c6e585/src/components/App/Importers/AddImporter.vue#L54-L56

we are saving the credential into loginFields element

PR #73, opened 2020-08-16
Edit global config in UI

baruchiro:
image
image

In forms, we have two ways- make the form reactive so any change will **immediately update** the state and the config, and the other way- **clone** the editable object, make the changes and update the config **only when the user press "Send"**.

I prefer the second way (but I know nowadays the first way come to be common).

In the second way, in Vue, we have two ways:
 - Split the two-way binding by setting the value of the field with the **original value**, but listen to the `change` event to update the **local component state**, and send the updated state when the user submits.
 - Clone **the whole object** (by `JSON.parse(JSON.stringify` for example) and **bind** it to the UI.

In this form I selected the second way.

### DIFF
--- a/src/components/App/ConfigEditor.vue
+++ b/src/components/App/ConfigEditor.vue
@@ -1,5 +1,8 @@
 <template>
-  <v-form ref="form" v-model="validated">
+  <v-form
+    ref="form"
+    v-model="validated"
+  >
     <v-container>
       <v-row justify="center">
         <v-col class="py-0">
@@ -11,7 +14,10 @@
             @input="updateConfig($event, 'numDaysBack')"
           />
         </v-col>
-        <v-col cols="auto" class="py-0">
+        <v-col
+          cols="auto"
+          class="py-0"
+        >
           <v-checkbox
             label="Show browser"
             :value="storeGlobalConfig.showBrowser"
@@ -21,7 +27,13 @@
       </v-row>
     </v-container>
 
-    <v-btn color="primary" :disabled="!readyToSave" @click="submitForm()">Save</v-btn>
+    <v-btn
+      color="primary"
+      :disabled="!readyToSave"
+      @click="submitForm()"
+    >
+      Save
+    </v-btn>
   </v-form>
 </template>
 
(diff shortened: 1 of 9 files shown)


src/components/App/ConfigEditor.vue
```diff
@@ -0,0 +1,90 @@
...
+      color="primary"
+      :disabled="!readyToSave"
+      @click="submitForm()"
+    >
+      Save
+    </v-btn>
+  </v-form>
+</template>
+
+<script lang="ts">
```
baruchiro, line 40: Note this is a TS component, it works.

src/components/App/ConfigEditor.vue
```diff
@@ -0,0 +1,90 @@
...
+import Vue from 'vue';
+import { mapGetters, mapActions } from 'vuex';
+import { GLOBAL_CONFIG_GETTER, UPDATE_GLOBAL_CONFIG_ACTION } from '@/store/modules/Config';
+import { VForm } from '@/types/vuetify';
```
baruchiro, line 44: Workaround for `Vue.$refs` that hold an abstract `Vue` instances so you have to convert it to the specific type.

src/originalBudgetTrackingApp/configManager/configManager.ts
```diff
@@ -16,10 +16,10 @@ export interface Config {
...
     ynab?: YnabConfig;
   };
-  scraping: {
-    'numDaysBack': number;
-    'showBrowser': boolean;
-    'accountsToScrape': AccountToScrapeConfig[];
+  accountsToScrape: AccountToScrapeConfig[];
+  globalConfig: {
+    numDaysBack: number;
+    showBrowser: boolean;
```
baruchiro, line 22: @brafdlog Split the `accountsToScrape` from the other global config.

baruchiro: The is a trade-off here. I want all the configs that configured by the `configEditor` in one sub-object, but it means that any different sub-object will need to take its config from the `globalConfig`.

And the opposite- you suggesting each config will be in its used object, and the `configEditor` will need to collect them to one view.

OK, we just need to take one of the options.

baruchiro: I mean, I have no objection to your proposal, I'm working on it.

PR #76, opened 2020-08-18
Event emitter

baruchiro:
We have work to do here.

In this code I just implemented the `eventEmitter` in some way, I have no experience and knowledge about conventions and best practices with `eventEmitters`, so we need to think about all. I just started to see if we can learn something about what we need to do.

And I learned.

We need to define the `event`s we want to `emit`. (I think we need to define the event types to know what the event is, and what data to expect, and what to do...)

Currently, I used four event types:
 - `status`: Generic "log" message, just reporting about step in our flow.
 - `progress`: The `israeli-bank-scrapers` expose only one generic method to register a listener (see [here](https://github.com/eshaham/israeli-bank-scrapers/blob/70d5d6d6b3cdad8967fc15ca8169b3ef3e25ce39/src/scrapers/base-scraper.ts#L192-L194)), so I used `progress` to different between `status` and `progress`.
- `error`: To error (but if we are throwing the error, do we need also to `emit`? I think my code is throwing the same error twice, because it `emit` and `throw`, then `catch` and `emit` and `throw` again)
- `finish`: We have some `finish` events, one for each scrape finished, and one for each output finished?

Also, I will write comments on the code about the decisions I make.


src/originalBudgetTrackingApp/commonTypes.ts
```diff
@@ -1,7 +1,31 @@
 import { Transaction } from '@/originalBudgetTrackingApp/bankScraper';
+import { CompanyTypes } from '@brafdlog/israeli-bank-scrapers-core';
+
+export interface ProgressEmitter {
+  onProgress(func: (...args: any[]) => void): void
+}
```
baruchiro, line 6: To expect [this method](https://github.com/eshaham/israeli-bank-scrapers/blob/83c4b664137e9bfdd76d8ef71cdce47a55733f58/src/scrapers/base-scraper.ts#L192-L194) in the `israeli-bank-scrapers`

src/originalBudgetTrackingApp/commonTypes.ts
```diff
@@ -1,7 +1,31 @@
...
+export interface ScrapingEventEmitter {
+  on<U extends keyof ScrapingEvents>(
+    event: U, listener: ScrapingEvents[U]
+  ): this;
+
+  emit<U extends keyof ScrapingEvents>(
+    event: U, ...args: Parameters<ScrapingEvents[U]>
+  ): boolean;
+
+  listenTo(emitter: ProgressEmitter): void
```
baruchiro, line 30: Start to listen to `israeli-bank-scrapers` events

src/originalBudgetTrackingApp/commonTypes.ts
```diff
@@ -1,7 +1,31 @@
...
 }
+
+export interface ScrapingEvents {
+  status: (status: string, data?: { name?: string, [other: string]: any }) => void,
+  progress: (...args: any[]) => void,
+  error: (message: string, data?: { companyId?: CompanyTypes, error?: Error, [other: string]: any }) => void,
+  finish: (data?: { companyId?: CompanyTypes, accountId?: string, [other: string]: any }) => void
+}
+
+export interface ScrapingEventEmitter {
```
baruchiro, line 21: I don't require a "real" event emitter (`require('events').EventEmitter`) because of `EventEmitter` is just an implementation of the `emit` method.
Actually now I think this interface even should not include the `on` method, because you need to decide if you really need this method. The scraping method doesn't use this method.
(See in the `ConsoleEmitter`, you don't have to use this method at all)

src/originalBudgetTrackingApp/index.ts
```diff
@@ -21,64 +22,66 @@ const DATE_FORMAT = 'DD/MM/YYYY';
 
 export const { getYnabAccountDetails } = ynab;
 
-export async function scrapeAndUpdateOutputVendors() {
+export async function scrapeAndUpdateOutputVendors(optionalEventEmitter?: ScrapingEventEmitter) {
+  const eventEmitter = new EmptyEventEmitterAdapter(optionalEventEmitter);
```
baruchiro, line 26: If you keep the `ScrapingEventEmitter` as `nullable`, so you have to pass it as `nullable` to all methods, and I was afraid the if some method declaring it as `nullable`, you may forget to move it within out internal implementation, and there will be no error.

To solve this I declared it as `nullable` only in the `export`ed method, then I adapted it to `EmptyEventEmitterAdapter`, and then I can pass it as not null, even if it null.

Not sure it is a good practice at all.

baruchiro:
@brafdlog I want you to review this, this is **draft** because we didn't discuss it, so I just POC it and we can discuss here or in a meeting.

brafdlog:
> @brafdlog I want you to review this, this is **draft** because we didn't discuss it, so I just POC it and we can discuss here or in a meeting.

I will review as soon as I can

baruchiro:
Hope I did the merge correctly

brafdlog (commented):
I know we discussed passing an event emitter to the main `scrapeAndUpdateOutputVendors` function but I now see that it complicates things. The fact that we get it as a parameter forces you to handle the case that it was not passed and to expose the fact of listening to the israeli-bank-scrapers event emitter which IMO should be an internal implementation detail.

I have two alternatives in mind:

1. Have a common module for the event emitter that both the lib code and the app code will import.
2. Make the `scrapeAndUpdateOutputVendors` create the event emitter and continue the flow of scraping without waiting for the promise to resolve - so you will get the event emitter immediately and can start listening on it and getting events. The advantage of this flow is that the lib code fully controls this and can hide any implementation details. The disadvantage is that the main function does not wait for the promise inside which makes it a bit less intuitive and forces calling code to work with the events to know when it is done.

Another thing is that I am not sure what is the best way to type the events additional potential options:

- https://stackoverflow.com/a/39145058/1391050
- https://github.com/sindresorhus/emittery

brafdlog:
And, I think our events should match the main stages of the flow and work nicely with the multiple accounts during import and export.

Something like:

```
IMPORT_PROCESS_START // Start the whole import process

IMPORTER_START // Start import of specific account

IMPORTER_PROGRESS

IMPORTER_ERROR

IMPORTER_END

IMPORT_PROCCESS_END

EXPORT_PROCESS_START

EXPORTER_START

EXPORTER_PROGRESS

EXPORTER_ERROR

EXPORTER_END

EXPORT_PROCESS_END
```

Where the `IMPORTER_*` and `EXPORTER_*` events are for importing/exporting of specific accounts so there such events will be thrown once for each importer and can even happen concurrently.

brafdlog:
Closing this PR because we have #99 for the same functionality

PR #79, opened 2020-08-26
Convert tests to TypeScript

baruchiro:


test/e2e/specs/Launch.spec.ts
```diff
@@ -73,9 +76,17 @@ const skip = true;
...
       const imgBuffer = await browserWindow.capturePage();
-      fs.writeFileSync(screenshotFile, imgBuffer);
+      fs.writeFileSync(screenshotFile, imgBuffer.toBitmap());
     }
   });
 
   afterAll(async () => stopServe());
 });
+
+declare global {
```
baruchiro, line 86: Workaround for `global`. Same as `shims.d.ts`. We need something better but for now, I did this.

brafdlog (approved):

PR #81, opened 2020-09-07
Create codeql-analysis.yml

brafdlog:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,62 @@
...
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
+
+    # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
+    # If this step fails, then you should remove it and run the build manually (see below)
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v1
```
baruchiro, line 48:
```suggestion
    # - name: Autobuild
    #  uses: github/codeql-action/autobuild@v1
```

This step is for compiled languages only.

See https://github.com/brafdlog/budget-tracking/runs/1080955027?check_suite_focus=true#step:5:26

baruchiro (approved):
I hope we have time to manage this, and that the result will not be just emails we ignore.

PR #88, opened 2020-09-07
Align casing of file and folder names resolves #57

### LINKED ISSUE #57: The consistent naming convention in the repo is camelCase for directories and files
The consistent naming convention in the repo is camelCase for directories and files so I would rename:

`config-manager` -> `configManager`
`default-config` -> `defaultConfig`

_Originally posted by @brafdlog in https://github.com/brafdlog/budget-tracking/pull/54#issuecomment-643321496_

I think the Vue part is not full compatible with this rule. But also need to think maybe in Vue the convention is different- https://vuejs.org/v2/style-guide/#Single-file-component-filename-casing-strongly-recommended

brafdlog:


brafdlog:
For #57

.github/workflows/release.yml
```diff
@@ -70,6 +70,7 @@ jobs:
         GOOGLE_CLIENT_ID: ${{ secrets.GOOGLE_CLIENT_ID }}
         GOOGLE_CLIENT_SECRET: ${{ secrets.GOOGLE_CLIENT_SECRET }}
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        NODE_OPTIONS: --max-old-space-size=4096
```
baruchiro, line 73: Did you have any problems?

baruchiro (approved):

.github/workflows/release.yml
brafdlog: Yes. The release crashed on out of memory

baruchiro: Interesting, what has changed?

brafdlog: Don't know... But the mac release is relatively new and oom issues can happen inconsistently if it gets close to the limit

baruchiro: Oh, Mac release, OK

PR #91, opened 2020-09-07
Extract logic from lib index file to dedicated modules - Resolves #90

### LINKED ISSUE #90: Break the lib index file into separate files for the major parts of the flow
Original comment:
I want to remove all the code from the root `index.ts`, to get a clear list about what the backend exporting. I have three (or more) points:

Move all the code from `index.ts` to separate files in a logical order. The **export** functionality should be in its file, the same with the **import** functionality, and the method that takes from **import** to **export** in another file. 
 In this way, we can validate (I hope there is lint for that) the `Vue` access only to the `@/originalBudgetTrackingApp`, and the backend itself access only to the [`flow`](https://github.com/brafdlog/budget-tracking/tree/cleanIndex/src/originalBudgetTrackingApp/flow) module, and this module uses its "internal" methods to import and export. 
 With this structure you can understand half of the code only by its module.

brafdlog:
Extract the logic from the lib index file into dedicated modules for the different parts of the flow.

Probably easier to review the changes commit by commit because of the folder structure changes.


baruchiro:
I'm trying to think about a way to force us to import only from `originalBudgetTrackingApp` and not from any `originalBudgetTrackingApp/*` when we are on the FrontEnd code.

Maybe this:
https://eslint.org/docs/rules/no-restricted-imports
Just keep in your mind

src/originalBudgetTrackingApp/index.ts
```diff
@@ -1,25 +1,16 @@
...
 export const { inputVendors } = bankScraper;
 
-type AccountToScrapeConfig = configManager.AccountToScrapeConfig;
-
-const TRANSACTION_STATUS_COMPLETED = 'completed';
-const DATE_FORMAT = 'DD/MM/YYYY';
-
-export const { getYnabAccountDetails } = ynab;
-
 export async function scrapeAndUpdateOutputVendors() {
```
baruchiro, line 14: OK, you can stay here, but if you're growth I will throw you to a different file!

src/originalBudgetTrackingApp/index.ts
```diff
@@ -1,25 +1,16 @@
...
   const config = await configManager.getConfig();
```
baruchiro, line 15: The `config` should be an argument of this function!
I don't know if to open an issue about the, or just solve it on the fly.

But I think it is an important idea- the `ConfigManager` should be decoupled from the `scraping`

src/originalBudgetTrackingApp/setupHelpers.ts
```diff
@@ -0,0 +1,40 @@
...
+  console.log('------YNAB Budgets------');
+  console.log(ynabAccountData.budgets);
+  console.log();
+  console.log('------YNAB Accounts------');
+  console.log(ynabAccountData.accounts);
+  console.log();
+  console.log('------Financial institutions account numbers------');
+  console.log(companyIdToAccountNumbers);
+  console.log();
+}
```
baruchiro, line 21: Why not in `ynab` module?

src/originalBudgetTrackingApp/setupHelpers.ts
```diff
@@ -0,0 +1,40 @@
...
+  console.log('Fetching data from financial institutions to determine the account numbers');
+  const companyIdToTransactions = await scrapeFinancialAccountsAndFetchTransactions(config, startDate);
+  const companyIdToAccountNumbers: Record<string, string[]> = {};
+  Object.keys(companyIdToTransactions).forEach((companyId) => {
+    let accountNumbers = companyIdToTransactions[companyId].map((transaction) => transaction.accountNumber);
+    accountNumbers = _.uniq(accountNumbers);
+    companyIdToAccountNumbers[companyId] = accountNumbers;
+  });
+  return companyIdToAccountNumbers;
+}
```
baruchiro, line 40: Why not in `bankScraper` module?

src/originalBudgetTrackingApp/setupHelpers.ts
brafdlog: I considered this. Since the setup of ynab requires matching account ids in ynab and the account ids in the financial and fetching all those ids is annoying, I wrote these to make it easier to do the setup. This is another flow of setup scripts (in addition to the flows of config and scrape) so I extracted it to a file of its own. 

I get your point that this is ynab specific so I will leave it in ynab and reexport it from here

brafdlog: On second thought I think we should keep it separated in the setup helpers. As you can see this is very specific code that is only relevant to the setup flow. It uses ynab's `getYnabAccountDetails` as a building block but the rest is not related to the ynab api and I don't want to mix this code with the regular flow.

WDYT?

src/originalBudgetTrackingApp/setupHelpers.ts
brafdlog: This I will move into the importTransactions module since it is related logic and uses the `scrapeFinancialAccountsAndFetchTransactions` function defined there

src/originalBudgetTrackingApp/index.ts
brafdlog: I will fix this in the config pr

brafdlog: Fixed in #92

brafdlog:
> I'm trying to think about a way to force us to import only from `originalBudgetTrackingApp` and not from any `originalBudgetTrackingApp/*` when we are on the FrontEnd code.
> 
> Maybe this:
> https://eslint.org/docs/rules/no-restricted-imports
> Just keep in your mind

Opened a separate issue for this #93

src/originalBudgetTrackingApp/setupHelpers.ts
baruchiro: If I understand you correctly, this is just an "external tool" to easily config the `ynab` at the first time (?).

If so, yes, let's keep it far from the library/module code. Even in a different folder, `tools` or something.

But to understand, if it not in the regular config+scrape flow, how can you run it? By running the code itself? How it published?

baruchiro: Do not pay too much attention to where I said put it. Maybe the file is in the right place.

baruchiro (approved):

src/originalBudgetTrackingApp/setupHelpers.ts
brafdlog: At this point it is running the code itself. We could create a ui flow for setup that will do this behind the scenes and instead of printing it to the console will help you fill the config. But at this stage it is just a code tool.

PR #99, opened 2020-09-13
Event emitter v2

brafdlog:
This is a work in progress


src/originalBudgetTrackingApp/eventEmitters/EventEmitter.ts
```diff
@@ -0,0 +1,76 @@
...
+  [EventNames.IMPORTER_END]: ImporterEndEvent
+  [EventNames.IMPORT_PROCESS_END]: { }
+  [EventNames.EXPORT_PROCESS_START]: { }
+  [EventNames.EXPORTER_START]: ExporterEvent
+  [EventNames.EXPORTER_PROGRESS]: ExporterEvent
+  [EventNames.EXPORTER_ERROR]: ExporterErrorEvent
+  [EventNames.EXPORTER_END]: ExporterEvent
+  [EventNames.GENERAL_ERROR]: ErrorEvent
+  [EventNames.LOG]: BudgetTrackingEvent
+};
```
baruchiro, line 70: I just found a better way to create an `enum` like this.
You can create your type, for example

```ts

type EventDataMap = {
  IMPORT_PROCESS_START: ImportProcessStartEvent
  IMPORTER_START: ImporterEvent
  ...
};
```

Then to export the keys as enum you can
```ts
export type EventNames = keyof EventDataMap
```

brafdlog: That was my initial thought as well. But the result of `keyof EventDataMap` is a union of all the strings of the keys. It does not behave like an enum and doesn't allow me to use the `EventNames.xxxxxxx` to get a specific value.
Screen Shot 2020-09-14 at 15 26 24

brafdlog:
@baruchiro Made a few more changes. The PR is now ready from my point of view.

baruchiro:
I will look on this as soon as I can, but if you can comment in places you think you can point on things we talked about, it will be good and will help me to understand the change.

src/originalBudgetTrackingApp/commonTypes.ts
```diff
@@ -12,8 +13,17 @@ export enum OutputVendorName {
   GOOGLE_SHEETS = 'googleSheets'
 }
 
+export type ExportTransactionsParams = {
+  transactionsToCreate: EnrichedTransaction[];
+  startDate: Date;
+  outputVendorsConfig: Config['outputVendors'];
+  eventEmitter: BudgetTrackingEventEmitter
+}
```
baruchiro, line 21:
```suggestion
export type ExportTransactionsParams = {
  transactionsToCreate: EnrichedTransaction[];
  startDate: Date;
  outputVendorsConfig: Config['outputVendors'];
}
```
I think the `eventEmitter` should be a different argument because it not related to the exporter itself. It is not part of the exporter input.

src/originalBudgetTrackingApp/commonTypes.ts
```diff
@@ -12,8 +13,17 @@ export enum OutputVendorName {
...
+
 export interface OutputVendor {
   name: OutputVendorName;
   init?: (outputVendorsConfig: Config['outputVendors']) => Promise<void>;
-  exportTransactions: (transactionsToCreate: EnrichedTransaction[], startDate: Date, outputVendorConfig: Config['outputVendors']) => Promise<any>;
+  exportTransactions: ({
+    transactionsToCreate, startDate, outputVendorsConfig, eventEmitter
+  }: ExportTransactionsParams) => Promise<any>;
```
baruchiro, line 28:
```suggestion
  exportTransactions: (exportParams: ExportTransactionsParams, eventEmitter: BudgetTrackingEventEmitter) => Promise<any>;
```

src/originalBudgetTrackingApp/eventEmitters/EventEmitter.ts
baruchiro: 😌I remember `keyof` return an `enum`, my mistake

src/originalBudgetTrackingApp/export/exportTransactions.ts
```diff
@@ -1,28 +1,42 @@
...
+  startDate: Date, eventEmitter: BudgetTrackingEventEmitter
 ) {
+  await eventEmitter.emit(EventNames.EXPORT_PROCESS_START);
   const executionResult = {};
   const allTransactions = _.flatten(Object.values(companyIdToTransactions));
 
   for (let j = 0; j < outputVendors.length; j++) {
     const outputVendor = outputVendors[j];
     if (outputVendorsConfig[outputVendor.name]?.active) {
+      const baseExporterEventData = { name: outputVendor.name.toString(), allTransactions };
```
baruchiro, line 19: I think we will prefer the least event emitter code in the flow as we can, right?
I'm just asking, what do you think about hiding the `emit(eventName, payload)` methods by implement methods like `emitExporterStart(name: string, transactions: EnrichedTransactions[])` so you can use one simple line `eventEmitter.emitExporterStart(outputVendor.name.toString(), allTransactions)`?

Of course, this requires a lot of specific code in the `BudgetTrackingEventEmitter`.

src/originalBudgetTrackingApp/export/exportTransactions.ts
```diff
@@ -1,28 +1,42 @@
...
+      await eventEmitter.emit(EventNames.EXPORTER_START, baseExporterEventData);
+      try {
+        const vendorResult = await outputVendor.exportTransactions({
+          transactionsToCreate: allTransactions, startDate, outputVendorsConfig, eventEmitter
+        });
+        await eventEmitter.emit(EventNames.EXPORTER_END, baseExporterEventData);
+        executionResult[outputVendor.name] = vendorResult;
+      } catch (e) {
+        await eventEmitter.emit(EventNames.EXPORTER_ERROR, { ...baseExporterEventData, error: e });
+        throw e;
```
baruchiro, line 31: Need to think/declare the error handling with the `eventEmitter`.
Since the error handling work with throwing up, it might lead to `emit` in each throwing level. In that case we can declare that the error events are just to emit an **info** and shouldn't affect the logic because we actually throw.

Another option is to throw custom error inside the code, so you can throw the `eventName` through all the levels and both `emit` and `throw` it out once, in the highest level, but with the specific eventName.

src/originalBudgetTrackingApp/export/exportTransactions.ts
baruchiro: I think also about setting the `name` of the exporter once so the `eventEmitter` will add it automatically in the related events, like **scopes** in some loggers.
But about this and the idea above, maybe I have good suggestions, but we want to merge this and all, so think carefully, maybe we can forego my suggestions for now.

src/originalBudgetTrackingApp/eventEmitters/EventEmitter.ts
```diff
@@ -0,0 +1,76 @@
...
+
+type EmptyEvents = EventNames.IMPORT_PROCESS_START | EventNames.IMPORT_PROCESS_END | EventNames.EXPORT_PROCESS_START | EventNames.EXPORT_PROCESS_END;
+
+export class BudgetTrackingEventEmitter extends Emittery.Typed<EventDataMap, EmptyEvents> {
+
+}
```
baruchiro, line 76: What about exporting interface for only `emit` methods?
Have you decided to give it up now?
It may be an edge case that we may to take later, as needed.

src/originalBudgetTrackingApp/import/importTransactions.ts
```diff
@@ -1,48 +1,57 @@
...
 
 type AccountToScrapeConfig = configManager.AccountToScrapeConfig;
+type Config = configManager.Config;
 type ScrapingConfig = Config['scraping'];
 
 const TRANSACTION_STATUS_COMPLETED = 'completed';
-const DATE_FORMAT = 'DD/MM/YYYY';
 
-export async function scrapeFinancialAccountsAndFetchTransactions(scrapingConfig: ScrapingConfig, startDate: Date) {
+export async function scrapeFinancialAccountsAndFetchTransactions(scrapingConfig: ScrapingConfig, startDate: Date, eventEmitter: BudgetTrackingEventEmitter) {
```
baruchiro, line 16:
```suggestion
export async function scrapeFinancialAccountsAndFetchTransactions(scrapingConfig: ScrapingConfig, startDate: Date, eventEmitter?: BudgetTrackingEventEmitter) {
```
Should be optional I think, or default one, if we don't want to handle null (although it is just to add `?` in `eventEmiter?.emit`

baruchiro: Sorry, ignore this. I see the `scrapeAndUpdateOutputVendors` function

src/originalBudgetTrackingApp/export/exportTransactions.ts
brafdlog: I was thinking about error handling as well. There are cases where we would want it to stop the flow and cases where we won't. For example, if one importer is failing, it is not necessarily a reason to crash the whole process. I would expect the other ones to complete successfully  and the one that fails to fail (which will be reflected in the ui of the progress once that exists).

So, in this case I would remove the throw from here all together and just emit the event (not doing this yet because since we don't have a status indication in the ui at the moment, if an errors isn't thrown it will look like it succeeded even though it failed).

What still looks like it doesn't make sense is that ynab emits and throws, and then the exporter emits and throws again.
The reason that code is there is because when I didn't have a catch around the ynab code, I didn't get the full details of the error that was thrown. This is probably the issue node has with async stack traces.

src/originalBudgetTrackingApp/export/exportTransactions.ts
brafdlog: Regarding the first comment. Its a good idea, but it requires one of two:

Either we accept the event name as a param, and then the calls look like this: 
```
await eventEmitter.emitExporterEvent({ eventName: EventNames.EXPORTER_START, outputVendor, allTransactions });
```

Which is not much better than the current situation, or as you suggested we add for every event name a function in the event emitter, which is quite allot of functions. So I think we should leave it as is for now

brafdlog: Regarding the second comment, that is also a good idea. I would like to have all this metadata sit in a scope or some other type of state instead of sending it as parameters in every emit call. The complexity here is that I would like the importers to run concurrently, and so maintaining scope that makes sense will probable require something like:

```
eventEmitter.runInScope(scopeParams, () => {
  // Code we want to run
}))
```

Makes the code a but strange, Don't think it is worth it at this point either.

Thoughts or ideas?

src/originalBudgetTrackingApp/eventEmitters/EventEmitter.ts
brafdlog: Forgot about it yesterday. Done

baruchiro (approved):
I need to read [this](https://github.com/brafdlog/budget-tracking/pull/99#discussion_r488592253) again, but approved

brafdlog:
> I need to read [this](https://github.com/brafdlog/budget-tracking/pull/99#discussion_r488592253) again, but approved

We can discuss it if needed. In any case I don't think it should block merging

src/originalBudgetTrackingApp/export/exportTransactions.ts
baruchiro: I think I understand. Not talking about specific implementation, in general we need to handle errors for each exporter and importer, and in high level I think we need to `emit` that error and **collect** it (for such `results` object we already have, I think).

If we encapsulate this specific case, we still have other optional errors, but I can guess it is more simple and we can only emit for these cases in the highest level.

Ok, it is not important how to implement this, it is just important to know about the potential issue and how we can possibly solve it.

> What still looks like it doesn't make sense is that ynab emits and throws, and then the exporter emits and throws again.
> The reason that code is there is because when I didn't have a catch around the ynab code, I didn't get the full details of the error that was thrown. This is probably the issue node has with async stack traces.

This is something I didn't understand, maybe because I don't familiar with the `ynab` code.

I think we need to open an issue about error handling and emitting, first to work on it, of course, but also to document our ideas and concern.

brafdlog: I created an issue for consistent error handling #105

PR #100, opened 2020-09-15
Update readme

### LINKED ISSUE #85: Update the readme and add to it the setup flow for the app from scratch
- Update the readme
- Document in Readme the up to date flow of setting up the app from scratch with importers and exporters

brafdlog:
Resolves  #85 

Once the configuration ui is in, I will update the steps of how to setup the project in the documentation.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -5,7 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
-    <title>israeli-bank-scrapers-desktop</title>
+    <title>Budget tracking</title>
     <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:100,300,400,500,700,900">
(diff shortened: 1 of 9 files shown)


public/index.html
```diff
@@ -5,7 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
-    <title>israeli-bank-scrapers-desktop</title>
+    <title>Budget tracking</title>
```
baruchiro, line 8: Do you want to think about less generic name?
The name from my repo was "Oshi"-"עו"שי".
Can't say I'm in love with this name, but I just think if we want another name, we need to decide now, before public release.

src/components/app/ReportProblemDialog.vue
```diff
@@ -107,7 +107,7 @@ ${detailes}` : '';
 ${log}
 \`\`\`` : '';
 
-  return 'https://github.com/baruchiro/israeli-bank-scrapers-desktop/issues/new?'
+  return 'https://github.com/brafdlog/budget-tracking/issues/new?'
```
baruchiro, line 110: See #103

PR #106, opened 2020-09-16
Propose more accurate EventEmitter

baruchiro:
We already talked about that, but why I was thinking about the `EventPublisher` type?

Because this interface is the required interface by our process. We don't need to use `EventEmitter` just becuase there is one. All we can do is to follow its part of the interface, to let who use our method to use `EventEmitter` easily if he want to.

Just imagine the `EventEmitter` idea- it is a stand-alone object to implement *Producer-Consumer*, so both the Producer and Consumer are accessing to it.

But here, we have more coupled case, and it's OK, because we don't need a the idea of the `EventEmitter`, we just need a *Callback*.

### DIFF
--- a/src/originalBudgetTrackingApp/eventEmitters/consoleEmitter.ts
+++ b/src/originalBudgetTrackingApp/eventEmitters/consoleEmitter.ts
@@ -1,11 +1,9 @@
 /* eslint-disable class-methods-use-this */
-import { BudgetTrackingEventEmitter, EventPublisher } from '@/originalBudgetTrackingApp/eventEmitters/EventEmitter';
+import { EventNames, EventPublisher } from '@/originalBudgetTrackingApp/eventEmitters/EventEmitter';
 
-export function buildConsoleEmitter(): EventPublisher {
-  const consoleEmitter = new BudgetTrackingEventEmitter();
-  consoleEmitter.onAny((eventName, eventData) => {
+export default class ConsoleEmitter implements EventPublisher {
+  async emit(eventName: EventNames, eventData?: any) {
     // eslint-disable-next-line no-console
     console.log(`${eventName}:`, eventData);
-  });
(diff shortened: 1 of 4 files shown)


baruchiro:
Actually, there is an advantage (maybe a big one) when providing a real `EventEmitter` from the `EventEmitter.ts` file. Because there you have an implemented interface, and you can just listen (`on`) to the events, **without creating a class for it**.

But still, I think providing an implemented `EventEmitter` is good just as helper, and keep the interface clean as in this PR.

src/originalBudgetTrackingApp/eventEmitters/consoleEmitter.ts
```diff
@@ -1,11 +1,9 @@
 /* eslint-disable class-methods-use-this */
-import { BudgetTrackingEventEmitter, EventPublisher } from '@/originalBudgetTrackingApp/eventEmitters/EventEmitter';
+import { EventNames, EventPublisher } from '@/originalBudgetTrackingApp/eventEmitters/EventEmitter';
 
-export function buildConsoleEmitter(): EventPublisher {
-  const consoleEmitter = new BudgetTrackingEventEmitter();
-  consoleEmitter.onAny((eventName, eventData) => {
+export default class ConsoleEmitter implements EventPublisher {
+  async emit(eventName: EventNames, eventData?: any) {
```
brafdlog, line 5: This is strange to me. The emit should just publish the event, not handle it. The previous implementation created a generic emitter, and just implemented onAny that logged to the console. It handled the console log on the consumer side, which seems to me to make more sense.

brafdlog (commented):
I don't understand the benefit of these changes.

In the existing code, before the changes, as a consumer all you had to do is `new BudgetTrackingEventEmitter()` and you got the emitter.

What is broken that needs to be fixed and is improved with these changes?

PR #107, opened 2020-09-21
Implement exporters config ui

### LINKED ISSUE #83: Add ui for configuring exporters

brafdlog:
Resolves #83


brafdlog:
@baruchiro I went over the changes, looks fine on my end. If you approve my changes I will merge

src/store/modules/config.ts
```diff
@@ -2,7 +2,6 @@ import { randomHex } from '@/modules/encryption/crypto';
 import {
   AccountToScrapeConfig, Config, OutputVendorsConfig, OutputVendorsNames
 } from '@/originalBudgetTrackingApp/configManager/configManager';
-import defaultConfig from '@/originalBudgetTrackingApp/configManager/defaultConfig';
```
baruchiro, line 3: So you don't have problems like 'cannot access to "something" of undefined' when you just start the application without saved config?

src/originalBudgetTrackingApp/commonTypes.ts
```diff
@@ -10,7 +10,8 @@ export interface EnrichedTransaction extends Transaction {
 
 export enum OutputVendorName {
   YNAB = 'ynab',
-  GOOGLE_SHEETS = 'googleSheets'
+  GOOGLE_SHEETS = 'googleSheets',
+  JSON = 'json'
```
baruchiro, line 14: I think we need to remove this enum and use the types from `configManager` because they are all created from one type definition:
```ts
export type OutputVendorsNames = keyof Config['outputVendors']
export type OutputVendorsConfigs = Exclude<Config['outputVendors'][OutputVendorsNames], undefined>
export type OutputVendorsConfig<T extends OutputVendorsNames> = Exclude<Config['outputVendors'][T], undefined>
```

src/components/shared/formValidations.ts
```diff
@@ -0,0 +1,7 @@
+export function required(value) {
+  return !!value || 'Required.';
+}
+
+export function positive(value: number) {
+  return value > 0 || 'Must be grater than 0';
+}
```
baruchiro, line 7: Very good, it's so required that I thought there was already.

src/components/app/exporters/exportersCommon.ts
```diff
@@ -0,0 +1,25 @@
...
+  const vForm = ref<VForm>();
+
+  const exporter = reactive(cloneDeep(store.getters.Config.getExporter(exporterName)));
+
+  const validated = ref(true);
+  const changed = ref(false);
+  const readyToSave = computed(() => validated && changed);
+  const submit = async () => {
+    if (vForm.value?.validate()) {
+      await store.dispatch.Config.updateExporter({ name: exporterName, exporter });
```
baruchiro, line 17: I used *Vuex Actions* becuase it lets us to do some logic related to the commit-dispatch.
But I think now, with TS, we can declare the type we expected for dispatch and we can skip the Action.

src/store/modules/config.ts
brafdlog: The `configManager.getConfig()` function has a fallback to the default config if there is no config file defined, so if the store already fetched the config from the config manager there should be no problem.

Since you have the splash screen until the store is loaded I think we are good.

src/components/app/exporters/exportersCommon.ts
brafdlog: I am missing some context here...

src/originalBudgetTrackingApp/commonTypes.ts
brafdlog: That is not possible as is, because you cannot use a type as a value. For that you need an enum. 

What i will do is move the enum into the configManager and use the enum values as the keys for the outputVendor config like so:

```
export interface Config {
  outputVendors: {
    [OutputVendorName.GOOGLE_SHEETS]?: GoogleSheetsConfig;
    [OutputVendorName.YNAB]?: YnabConfig;
    [OutputVendorName.JSON]?: JsonConfig;
  };
...
```

baruchiro: OK then

src/components/app/exporters/exportersCommon.ts
baruchiro: *Vuex* have `getters`, `actions` and `commits`.
You can change the state only by `commit`, but if you want to do some logic, especially if it is an async logic, you need to do that with `action`.
I used `actions` to get a more comfort interface for changing the state, but the right way, if we don't have special logic, to do the change directly with `commit` (AKA `dispatch`)

src/components/app/exporters/YnabAccountMappingTable.vue
```diff
@@ -0,0 +1,81 @@
...
+      },
+      addAccountMapping() {
+        // @ts-ignore
+        this.$emit('addAccountMapping');
+      },
+      rules: {
+        required
+      }
+    };
+  }
```
baruchiro, line 75:
```suggestion
  setup(_, { emit }) {
    return {
      deleteAccountMapping(accountNumberToDelete) {
        emit('deleteAccountMapping', accountNumberToDelete);
      },
      addAccountMapping() {
        emit('addAccountMapping');
      },
      rules: {
        required
      }
    };
  }
```

src/components/app/exporters/YnabAccountMappingTable.vue
```diff
@@ -0,0 +1,81 @@
...
+      type: Array,
+      required: true
+    }
+  },
+  setup(_props) {
+    return {
+      deleteAccountMapping(accountNumberToDelete) {
+        // @ts-ignore
+        this.$emit('deleteAccountMapping', accountNumberToDelete);
+      },
```
baruchiro, line 66: You have a bug here:

1. Add new *Account to Ynab* item.
2. Set value for `Account number`.
3. Try to delete this item.

Also, the + button is stuck after that (and maybe without these steps)

searchlist2

baruchiro: I don't research this deeply. Just a question (I'm not sure how *Two Way Binding* works, I'm always googling it), why you're using these `events` and not just use `v-model`?

As I remember, `v-model` is shortcut for `value` prop and `input` event, so you can get an array of pairs, and add/remove/change it inside `YnabAccountMappingTable`, and `emit` it in any change.

src/components/app/exporters/YnabExporter.vue
```diff
@@ -0,0 +1,112 @@
+<template>
+  <v-form
+    ref="vForm"
+    v-model="validated"
+  >
+    <v-checkbox
+      v-model="exporter.active"
+      label="Active"
+      hide-details="auto"
+      @change="changed = true"
```
baruchiro, line 10:
```suggestion
      label="Active"
      @change="changed = true"
```
image

src/components/app/exporters/YnabExporter.vue
```diff
@@ -14,33 +14,94 @@
...
+        const updatedAccountNumbersToYnabAccountIds = {};
+        accountNumbersToYnabAccountIdsArray.value.forEach(({ accountNumber, ynabAccountId }) => {
+          updatedAccountNumbersToYnabAccountIds[accountNumber] = ynabAccountId;
+        });
+        ynabConfig.options.accountNumbersToYnabAccountIds = updatedAccountNumbersToYnabAccountIds;
+
+        dataToReturn.submit();
+      },
+      deleteAccountMapping: (accountNumber) => {
+        Vue.delete(ynabConfig.options.accountNumbersToYnabAccountIds, accountNumber);
```
baruchiro, line 96: This `Vue.set` and `Vue.delete` seems like a very bad limitation in Vue! 😥

src/components/app/exporters/YnabExporter.vue
```diff
@@ -14,33 +14,94 @@
...
+        dataToReturn.changed.value = true;
+      },
+      addAccountMapping: () => {
+        Vue.set(ynabConfig.options.accountNumbersToYnabAccountIds, '###', '###');
+      },
```
baruchiro, line 101:
```suggestion
      addAccountMapping: () => {
        Vue.set(ynabConfig.options.accountNumbersToYnabAccountIds, '', '');
      },
```

I prefer that way, you can disagree with me.

For now it's OK to work with `object`, but I think, at least in the form, it should be an array of `key-value`, with rule to not duplicate the keys. Because as it now, you have a problem with adding more then one items. You have to add an item, fill it, then add another one... But it is not critical for now.

baruchiro (approved):

src/components/app/exporters/YnabAccountMappingTable.vue
baruchiro: Solved. I don't know if my solution is better in all the aspects, but I think we need to use [`v-model`](https://vuejs.org/v2/guide/components.html#Using-v-model-on-Components).

Then I had a lot of problems, I worked on this a day, so first I managed to get it to work, and now we have something to improve.

src/components/app/exporters/YnabAccountMappingTable.vue
```diff
@@ -47,28 +47,48 @@
 </template>
 
 <script lang="ts">
-import Vue from 'vue';
+import Vue, { PropType } from 'vue';
 import { required } from '@/components/shared/formValidations';
+import { ref } from '@vue/composition-api';
+import { cloneDeep } from 'lodash';
+
+const objToArrayOfKeyValueArray = (obj: Record<string, string>) => Object.keys(obj).map((key) => ({ key, value: obj[key] }));
```
brafdlog, line 59: This code seems way more generic and unclear than it should be.

This is a very specific use case with specific values. The generic names and types here just make it harder to understand what is going on.

PR #108, opened 2020-09-21
Change exporters and importers ui idea

brafdlog:
I am thinking about changing the ui of importers and exporters from expandable panels to chips with icons that are clickable. Clicking on them will open the config in the modal.

These chips can also have an icon to indicate progress status.

This is an initial step in that direction. Still very ugly, but would like to get your feedback at this stage.

Screen Shot 2020-09-22 at 0 26 38
Screen Shot 2020-09-22 at 0 26 46
Screen Shot 2020-09-22 at 0 27 03


baruchiro:
@brafdlog Do you want to close this?

PR #110, opened 2020-10-04
Log events

baruchiro:
image

Using the `EventEmitter`, I'm printing the events to the UI, to see the process and errors if any.

It's very hard to work with the current `EventEmitter` design. You have to declare manually all the `EventNames`, and you don't have type-checking for that.
You can achieve type-checking, of course, but it creates a very complicated code.
I think we need to define the events as classes instead of interfaces, and yes, it will be noises in the backend, so maybe create a helper method to create these objects.

Anyway, don't know. We need more helper methods here.


src/components/app/MainContent.vue
```diff
@@ -3,68 +3,59 @@
...
-  name: 'MainContent',
+export default Vue.extend({
   components: {
-    LogLines, ConfigEditor
+    LogViewer, ConfigEditor
   },
-  data() {
+  setup() {
+    const inProgress = ref(false);
+    const succeeded = ref('unknown' as keyof typeof colors);
```
brafdlog, line 43: This doesn't ever change. Don't think this was what you intended.
Need to add back the logic that updates according to result of scraping

brafdlog: In general I think we should replace succeeded and in progress with a single status. Something like:
status: NOT_STARTED | IN_PROGRESS | SUCCESS | FAILURE

Then colors becomes STATUS_COLORS: Record<ScrapingStatus, string | null>

brafdlog: You can actually use the event emitter for setting the status instead of awaiting the promise (not a must):

eventPublisher.on(EventNames.IMPORT_PROCESS_START, () => {
    // IN_PROGRESS
  });
  
  eventPublisher.on(EventNames.IMPORT_PROCESS_END, () => {
    // SUCCESS
  });
  
  eventPublisher.on(EventNames.GENERAL_ERROR, () => {
    // ERROR
  });

src/components/app/LogsEventEmitter.ts
```diff
@@ -0,0 +1,21 @@
+import { EventEmitter } from '@/originalBudgetTrackingApp';
+import { EventNames } from '@/originalBudgetTrackingApp/eventEmitters/EventEmitter';
+import { Levels, LogEntry } from '../shared/log/types';
+
+export default (callback: (entry: LogEntry) => void) => {
+  const eventPublisher = new EventEmitter.BudgetTrackingEventEmitter();
```
brafdlog, line 6: I think it would be cleaner to add the level to the BudgetTrackingEvent itself, setting it by default to INFO and having error set it to ERROR. Then you won't need this class and can use the event publisher directly.

In the current implementation you don't have access to the full api of the emitter which is useful (see comment about the option to set the status by listening to events)

brafdlog: Also, we are losing data here that doesn't come from the message. For example, it would make sense in the future to display the icon of the account from which this log entry came from,for that you would need the account details but here they are filtered out.

src/originalBudgetTrackingApp/eventEmitters/EventEmitter.ts
```diff
@@ -19,14 +19,15 @@ export enum EventNames {
...
 
-interface ErrorEvent {
-  error: Error
-}
-
 interface BudgetTrackingEvent {
   message?: string;
 }
 
+// TODO: Maybe extend Error?
```
brafdlog, line 26: Why do we need to extend error? It has an error inside. This is an event so we want it to behave like other events. So I think it should extend BudgetTrackingEvent

brafdlog:
> image
> 
> Using the `EventEmitter`, I'm printing the events to the UI, to see the process and errors if any.
> 
> It's very hard to work with the current `EventEmitter` design. You have to declare manually all the `EventNames`, and you don't have type-checking for that.
> You can achieve type-checking, of course, but it creates a very complicated code.
> I think we need to define the events as classes instead of interfaces, and yes, it will be noises in the backend, so maybe create a helper method to create these objects.
> 
> Anyway, don't know. We need more helper methods here.

I don't fully understand what you mean. You could either list onAny if the use case is very generic (as it is in the current implementation of the logger). In this case (as an any generic code) you can only assume that the type is `BudgetTrackingEvent` which doesn't give you much, but you are listening on everything. Alternatively if your use case is more specific - which I think will happen in the next iterations of implementing the progress indication you can listen for specific events and get them fully typed.

I am not saying this implementation is optimal, I am just not sure what are the actual pain points so we can improve the current design. We can discuss this on the phone if you want.

src/components/app/LogsEventEmitter.ts
baruchiro: You're right, but I think you starting to understand the difficulty. We have so many event types, and if you want to get the account, for example, you have to explicitly say which events are related to account.

I think it is because we are using interfaces, so there is no meaning of the inheritance, actually. With interfaces, you just declare the shape of the object, but you can't know in the listener if the payload is an instance of something.

baruchiro:
I can't tell you exactly what is the pain. Of course, I can listen to a specific event to get the specific type. But I want it to be generic.

Also, take the `message` as an example- I expect all the events to have a message generated by the other event fields, I think if we will use classes, we can create the `message` as a function.

In short, I think the better is to let you try to improve my code, if you want and if you have time, and see if it hurts you.

src/components/app/LogsEventEmitter.ts
brafdlog: Changing to classes makes sense.

Maybe there are too many event types, but I would spend time on improving that together with a use cases that will use the specific events (like the progress indication)

brafdlog:
> I can't tell you exactly what is the pain. Of course, I can listen to a specific event to get the specific type. But I want it to be generic.
> 
> Also, take the `message` as an example- I expect all the events to have a message generated by the other event fields, I think if we will use classes, we can create the `message` as a function.
> 
> In short, I think the better is to let you try to improve my code, if you want and if you have time, and see if it hurts you.

Thats fine. As I said we will have a real use case once we implement the progress indication. For now I suggest just the changes in the comments

src/components/app/MainContent.vue
baruchiro: > In general I think we should replace succeeded and in progress with a single status. Something like:
> status: NOT_STARTED | IN_PROGRESS | SUCCESS | FAILURE
> 
> Then colors becomes STATUS_COLORS: Record<ScrapingStatus, string | null>

I think I don't understand... You just suggesting to change the `unknown | success | fail` with these `NOT_STARTED | IN_PROGRESS | SUCCESS | FAILURE`?

baruchiro: > You can actually use the event emitter for setting the status instead of awaiting the promise (not a must):
> 
> eventPublisher.on(EventNames.IMPORT_PROCESS_START, () => {
> // IN_PROGRESS
> });
> 
> eventPublisher.on(EventNames.IMPORT_PROCESS_END, () => {
> // SUCCESS
> });
> 
> eventPublisher.on(EventNames.GENERAL_ERROR, () => {
> // ERROR
> });

As I said before, I don't like to use `EventEmitter` as a "serial" flow, because I think it is less readable.

src/components/app/LogsEventEmitter.ts
baruchiro: > I think it would be cleaner to add the level to the BudgetTrackingEvent itself, setting it by default to INFO and having error set it to ERROR. Then you won't need this class and can use the event publisher directly.
> 
> In the current implementation you don't have access to the full api of the emitter which is useful (see comment about the option to set the status by listening to events)

https://github.com/brafdlog/budget-tracking/issues/113

baruchiro: OK then, let's keep it that way and we will improve it as needed as you suggested

src/components/app/MainContent.vue
brafdlog: Moved this discussion to the phone. Easier to communicate that way 🙂

brafdlog: Discussed on the phone. Easier to communicate that way 🙂

baruchiro: I did it. I started with `enum` but I think for now we can just use the `keyof typeof statusToColor`

src/originalBudgetTrackingApp/eventEmitters/EventEmitter.ts
baruchiro: I thought about that I think because then you getting the `Error.message`, but nevermind, we decided to continue to work to improve these types.

baruchiro:
@brafdlog Ready, waiting for your approval.

src/components/app/MainContent.vue
```diff
@@ -28,27 +28,30 @@ import ConfigEditor from './ConfigEditor.vue';
 import { LogEntry } from '../shared/log/types';
 import LogsEventEmitter from './LogsEventEmitter';
 
-const colors = {
-  success: 'green',
-  failed: 'red',
-  unknown: null
+const statusToColor = {
```
brafdlog, line 31: I think an enum is cleaner, instead of using the string constants every where you need to reference it. But up to you. Approved

brafdlog (approved):

src/components/app/MainContent.vue
baruchiro: I removed to `enum` because you always have to write `enumName.ENUM_VALUE`.

If it wasn't TS I would never use the string constant. But they are not just a string, because if you put a wrong string it will be a compilation issue.

PR #112, opened 2020-10-08
Enable JsonExporter

baruchiro:
Some fixes to JsonExporter. I'm using this exporter, also to easily test the flow because this is the simplest exporter.


src/modules/spreadsheet/spreadsheet.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
baruchiro, line 1: I deleted old `.js` files, so I don't want to try to fix this unused file. Anyway we will work on this file soon

src/originalBudgetTrackingApp/export/outputVendors/index.test.js
```diff
@@ -1,33 +0,0 @@
-/* eslint-disable global-require, import/no-dynamic-require */
```
baruchiro, line 1: This file was about test the `outputVendor` interface. We don't need this with TS

src/originalBudgetTrackingApp/import/importTransactions.ts
```diff
@@ -120,9 +121,3 @@ function transactionsDateComparator(t1: Transaction, t2: Transaction) {
   }
   return 1;
 }
-
-export function calculateTransactionHash({
-  date, chargedAmount, description, memo
-}: Transaction, companyId: string, accountNumber: string) {
-  return `${date}_${chargedAmount}_${description}_${memo}_${companyId}_${accountNumber}`;
-}
```
baruchiro, line 128: I moved this function (and maybe more functions in the future) to a separate file called `transactions/transactions`, this is something like "*Transactions Utils*"- helper methods to work with transactions (because I needed this function also in `JsonExporter`).

In addition, I removed the `companyId` and `accountNumber` from the hash, because we need the hash to easily identify transactions **no matter they come from**, and if we want to merge the imported transactions with existed transactions in the output vendor, we need to recalculate the hash for transactions in the output.
In short, I think we need to calculate the hash from the `Transaction`/`EnrichrdTransaction` itself. (So maybe add  `companyId` and `accountNumber` to the interface.

@brafdlog WDYT?

src/originalBudgetTrackingApp/transactions/transactions.ts
```diff
@@ -0,0 +1,114 @@
...
+//     title: 'Status',
+//   },
+// ];
+
+const transactionArrayToObject = (transactions: EnrichedTransaction[]) => transactions.reduce((acc, enrichedTransaction) => {
+  acc[enrichedTransaction.hash] = enrichedTransaction;
+  return acc;
+}, {} as Record<string, EnrichedTransaction>);
+
+// TODO removed company and account, need attention
```
baruchiro, line 95: See https://github.com/brafdlog/budget-tracking/pull/112/files#r501438587
Please do not resolve until I removing this `TODO`

src/originalBudgetTrackingApp/transactions/dates.ts
```diff
@@ -0,0 +1,5 @@
+export const formatDate = (date: Date) => `${date.getDate()}/${date.getMonth()}/${date.getFullYear()} ${date.getHours()}:${date.getMinutes()}`;
+
+export const compareObjects = (a: { date: string }, b: { date: string }) => compareStrings(a.date, b.date);
+export const compareStrings = (a: string, b: string) => compare(new Date(a), new Date(b));
+export const compare = (a: Date, b: Date) => a.getTime() - b.getTime();
```
brafdlog, line 5: compareStrings and compare are internal functions so I don't think they should be exported. And names could be clearer.


```suggestion
export const compareObjectsByDate = (a: { date: string }, b: { date: string }) => compareDateStrings(a.date, b.date);
const compareDateStrings = (a: string, b: string) => compareDates(new Date(a), new Date(b));
const compareDates = (a: Date, b: Date) => a.getTime() - b.getTime();
```

src/originalBudgetTrackingApp/export/outputVendors/json/json.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  EnrichedTransaction, ExportTransactionsFunction, OutputVendor, OutputVendorName
+} from '@/originalBudgetTrackingApp/commonTypes';
+import { mergeTransactions, sortByDate } from '@/originalBudgetTrackingApp/transactions/transactions';
+import { promises as fs } from 'fs';
+
+const parseTransactionsFile = async (filename: string) => fs.readFile(filename, { encoding: 'utf8' })
+  .then((content) => JSON.parse(content) as EnrichedTransaction[])
```
brafdlog, line 8: Why not use async await? The vast majority of the code uses it.

IMO it is clearer syntax.

brafdlog (commented):
Nice. see comments

src/originalBudgetTrackingApp/export/outputVendors/json/json.ts
baruchiro: I like the `then` syntax when we have a **pipeline**, you know, you just getting the result of one Promise to another Promise or permutation. I even think in this case it cleaner because the context of each Promise result is just the next `then`.

But of course I need to think again if I'm doing this just in the right places.

Anyway, I converted to `async` `await`. Push soon.

Soory if I have more the ususal English issues, I'm working with the *Github VSCode Extension* so no *Grammarly* at all 😂

src/originalBudgetTrackingApp/import/importTransactions.ts
baruchiro: OK, sounds right.

src/originalBudgetTrackingApp/transactions/transactions.ts
```diff
@@ -0,0 +1,30 @@
...
+export const calculateTransactionHash = ({
+  date, chargedAmount, description, memo
+}: Transaction) => {
+  return `${date}_${chargedAmount}_${description}_${memo}`;
+};
+
+export const mergeTransactions = (a: EnrichedTransaction[], b: EnrichedTransaction[]) => {
+  const aObj = transactionArrayToObject(a);
```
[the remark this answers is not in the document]
baruchiro: I'm taking into account that in JSON, for example (and in all "manual" exporters like "sheets" and CSV), the user may add some personal data into our DTOs- expanding our output.

But I think you're right, after outputting the transaction once, we don't need to support the case that this transaction changed **in the original source (the bank)**. (Am I right? What about "pending transactions"? Maybe in that case we need to update the already exported transaction?)

src/originalBudgetTrackingApp/transactions/dates.ts
```diff
@@ -0,0 +1,5 @@
+export const formatDate = (date: Date) => `${date.getDate()}/${date.getMonth()}/${date.getFullYear()} ${date.getHours()}:${date.getMinutes()}`;
```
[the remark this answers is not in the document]
baruchiro: Strange, maybe it was used in `.js` file. I think we will use it since I want to reduce the usage of the deprecated `moment` library, so sometime we will add this again, this is nothing.

brafdlog (approved):

src/originalBudgetTrackingApp/transactions/transactions.ts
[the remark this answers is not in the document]
brafdlog: I think we can juat skip pending transactions and thus only final data will be handled.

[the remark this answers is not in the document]
baruchiro: It depends on the user selection for my opinion.

PR #117, opened 2020-11-04
Removed credits from bottom bar

giladtaase:
Removed the developer credits from the bottom blue bar.
I left intact the link to github on the bottom right corner. If you'd like me to remove it (or the entire blue bar), please let me know.


baruchiro (approved):
Approved.

@brafdlog What about the link to Github and the whole bottom bar?

brafdlog (approved):
I think we can leave the bottom bar until we will implement a redesign

PR #119, opened 2020-11-08
Fixing UI bug in Ynab exporter: changes to account/ynab were not save…

giladtaase:
…d properly to the model


baruchiro (approved):
Consider rename the `keyVal` to something less confusing, for example, `obj` or `accountAndYnab`

PR #123, opened 2020-11-18
CSV exporter

giladtaase:


src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
```diff
@@ -0,0 +1,107 @@
...
+import { promises as fs } from 'fs';
+
+const stringify = require('csv-stringify/lib/sync');
+const parse = require('csv-parse/lib/sync');
+
+export function parseTransactions(csvText: string) {
+  return parse(csvText, {
+    columns: true,
+    cast: (value, context) => {
+      switch (context.column) {
```
baruchiro, line 15: - [ ] You have about 4 levels of indentations here, my suggestion is to move the `cast` to another function called something like `parseValue(value, columnName)`.

Also, personally most of the time I prefer to use an `object` as mapping, instead of a `switch` statement. But I can't explain to you how because the code is complicated.

- [ ] So another suggestion is also moving the conversions functions (inside the `cases`) to a separated function. When you're moving code to function, the function name can be the title of the code, and the code will be more understandable.

After these two suggestions I will iterate this file again, OK?

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
```diff
@@ -0,0 +1,107 @@
...
+            return value;
+          }
+        default:
+          return value;
+      }
+    }
+  }) as EnrichedTransaction[];
+}
+
+export const serializeTransactions = (transactions: EnrichedTransaction[]) => {
```
baruchiro, line 49: `export` is to make this method accessible outside the file, by assigning it as a property of the module. For example:

```nodejs
import myModule from 'myFile'
myModule.serializeTransactions()
```

But you are not using this function outside the file.

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export function parseTransactions(csvText: string) {
+  return parse(csvText, {
+    columns: true,
+    cast: (value, context) => {
+      switch (context.column) {
+        // note: we check here some number-type values.
+        // if there are undefined string-type values (memo?, for example) they'll be serialized as empty string.
+        // it means that the deserialized object will not be 100% identical to the serialzed.
+        // I'm not sure how to represent empty string differently than undefined string.
```
baruchiro, line 19: If you had tests on these cases, I would understand more what you expect.

(you have tests, but the tests are not including the cases of empty string or undefined, and I think in this case it is easy to understand your solution from test examples, instead of trying to read the code.)

test/unit/specs/csv.spec.ts
```diff
@@ -0,0 +1,44 @@
+import { TransactionStatuses, TransactionTypes } from '@brafdlog/israeli-bank-scrapers-core/lib/transactions';
+import { EnrichedTransaction } from '../../../src/originalBudgetTrackingApp/commonTypes';
+import { serializeTransactions, parseTransactions } from '../../../src/originalBudgetTrackingApp/export/outputVendors/csv/csv';
```
baruchiro, line 3:
```suggestion
import { EnrichedTransaction } from '@/originalBudgetTrackingApp/commonTypes';
import { serializeTransactions, parseTransactions } from '@/originalBudgetTrackingApp/export/outputVendors/csv/csv';
```

baruchiro (changes requested):
Thanks! good work, and after each PR you're starting to understand NodeJS and our principals more and more, keep going!

test/unit/specs/csv.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+      category: 'Finance',
+      hash: '11KJFLDKJ22__3231',
+      status: TransactionStatuses.Pending,
+      type: TransactionTypes.Installments
+    }
+  ];
+
+  const serialized = serializeTransactions(transactions);
+  const parsed = parseTransactions(serialized);
+  expect(parsed).toEqual(transactions);
```
brafdlog, line 43: Since the equality of these is not the most important thing in my opinion, I suggest using snapshot testing here for basic test cases:

`expect(parsed).toMatchSnapshot()`

And then add specific test cases for handling of empty strings, instalments etc

brafdlog: A good way to structure this is using describe blocks. Something like:

```
describe('CSV exporter', () => {
  describe('Serialize and parse transactions', () => {
    test('Basic test case', () => {
      ...
      expect().toMatchSnapshot();
    });
    test('should convert undefined to empty string', () => {
      
    });
    test('should parse instalments correctly', () => {
      
    });
  });
});
```

brafdlog (commented):
Nice work 👍 .

I added a few comments to Baruch's review.

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
```diff
@@ -0,0 +1,107 @@
...
+async function writeCsvFile(filePath: string, csvText: string) {
+  await fs.writeFile(filePath, csvText);
+}
+
+const exportTransactions: ExportTransactionsFunction = async ({ transactionsToCreate, outputVendorsConfig }) => {
+  const { filePath } = outputVendorsConfig.csv!.options;
+  const savedTransactions = await parseTransactionsFile(filePath);
+  const mergedTransactions = mergeTransactions(savedTransactions, transactionsToCreate);
+  const sorted = sortByDate(mergedTransactions);
+  writeCsvFile(filePath, serializeTransactions(sorted));
```
brafdlog, line 101:
```suggestion
  await writeCsvFile(filePath, serializeTransactions(sorted));
```

src/originalBudgetTrackingApp/configManager/defaultConfig.ts
```diff
@@ -7,8 +7,14 @@ const DEFAULT_CONFIG: Config = {
     accountsToScrape: [],
   },
   outputVendors: {
-    json: {
+    csv: {
       active: true,
```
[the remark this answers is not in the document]
giladtaase: Keeping CSV as default, let me know if you still want to change this

src/components/app/Exporters.vue
```diff
@@ -3,6 +3,9 @@
     <v-toolbar>
       <v-toolbar-title>Exporters</v-toolbar-title>
     </v-toolbar>
+    <expansion-panel title="Csv">
```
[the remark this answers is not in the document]
giladtaase: ok

src/components/app/exporters/CsvExporter.vue
```diff
@@ -0,0 +1,48 @@
...
+      ...setupExporterConfigForm(OutputVendorName.CSV),
+      rules: {
+        legalPath
+      }
+    };
+  }
+});
+
+</script>
+<style scoped/>
```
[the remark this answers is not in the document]
giladtaase: ok

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
```diff
@@ -0,0 +1,107 @@
+import {
+  EnrichedTransaction, ExportTransactionsFunction, OutputVendor, OutputVendorName
+} from '@/originalBudgetTrackingApp/commonTypes';
+import { mergeTransactions, sortByDate } from '@/originalBudgetTrackingApp/transactions/transactions';
+import { TransactionInstallments } from '@brafdlog/israeli-bank-scrapers-core/lib/transactions';
+import { promises as fs } from 'fs';
+
+const stringify = require('csv-stringify/lib/sync');
+const parse = require('csv-parse/lib/sync');
```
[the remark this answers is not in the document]
giladtaase: ok

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
giladtaase: ok

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
giladtaase: I'm using it in the test file test/unit/specs/csv.spec.ts
Is there a way of testing it without exposing it with export?

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
giladtaase: Ok, I will remove from the main code and add a short explanation in the test.

test/unit/specs/csv.spec.ts
giladtaase: Ok.
BTW for some reason VS code IDE doesn't recognize the file when I follow this format in the unit tests (though it works for the main code). The tests compile and run fine, but it's a bit of a pain working from the IDE.
Note that the other unit tests also follow that long format, I don't know if that's the reason.

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
```diff
@@ -0,0 +1,107 @@
...
+        // it means that the deserialized object will not be 100% identical to the serialzed.
+        // I'm not sure how to represent empty string differently than undefined string.
+        case 'chargedAmount':
+        case 'identifier':
+        case 'originalAmount':
+          if (value === '') {
+            return undefined;
+          }
+          return parseFloat(value);
+        case 'installments':
```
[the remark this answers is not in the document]
giladtaase: ok

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
```diff
@@ -0,0 +1,107 @@
...
+      { key: 'memo' },
+      { key: 'originalAmount' },
+      { key: 'originalCurrency' },
+      { key: 'processedDate' },
+      { key: 'accountNumber' },
+      { key: 'category' },
+      { key: 'hash' },
+      { key: 'status' },
+      { key: 'type' },
+      { key: 'category' }
```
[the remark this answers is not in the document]
giladtaase: No, a typo

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+    throw err;
+  }
+};
+
+async function writeCsvFile(filePath: string, csvText: string) {
+  await fs.writeFile(filePath, csvText);
+}
+
+const exportTransactions: ExportTransactionsFunction = async ({ transactionsToCreate, outputVendorsConfig }) => {
```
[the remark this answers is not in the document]
giladtaase: Habits of a lazy C programmer, if you put the helper function on top you don't have to declare it ;-)
Changed to match your style for consistency.

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
giladtaase: ok

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
giladtaase: @brafdlog Baruch is in Miluim, maybe you have an idea?

brafdlog: You can leave the export if you do test the function.

In general, we could use something like rewire for this but I don't see a need for that at this stage.

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
```diff
@@ -0,0 +1,107 @@
...
+      { key: 'originalAmount' },
+      { key: 'originalCurrency' },
+      { key: 'processedDate' },
+      { key: 'accountNumber' },
+      { key: 'category' },
+      { key: 'hash' },
+      { key: 'status' },
+      { key: 'type' },
+      { key: 'category' }
+    ],
```
brafdlog, line 68: I suggest changing the order to have the most important columns first
```suggestion
    columns: [
      { key: 'date' },
      { key: 'description' },
      { key: 'chargedAmount' },
      { key: 'memo' },
      { key: 'category' },
      { key: 'accountNumber' },
      { key: 'installments' },
      { key: 'originalAmount' },
      { key: 'originalCurrency' },
      { key: 'processedDate' },
      { key: 'identifier' },
      { key: 'hash' },
      { key: 'status' },
      { key: 'type' }
    ],
```

giladtaase: ok

src/originalBudgetTrackingApp/export/outputVendors/csv/csv.ts
giladtaase: ok

test/unit/specs/csv.spec.ts
giladtaase: ok

giladtaase:
@brafdlog Could you please review again?

brafdlog (approved):

Issue #120, opened 2020-11-09
Sheets Exporter: Support new Spreadsheet, new sheet, exists data


baruchiro:
After merging #118, need to validate the exporting to Google Spreadsheets code, includes

- [ ] Create a new Spreadsheet by name.
- [ ] Create a new sheet by const name.
- [ ] Read existed data from the sheet and merge with the new data.

baruchiro:
On second thought, creating a new Spreadsheet by name should be implemented in the UI context, since we have two different areas in the app- the config editor (the UI) and the scraping that gets the config and doing the job.

If we are putting a new spreadsheet name in the config, then passing the config as an argument to the scraping, the scraping will create a new spreadsheet, but the config still defines the Spreadsheet name.

I mean, during the scraping, the SpreadsheetId value will change from the new Spreadsheet name to its SpreadsheetId, but we are not changing the config during the scraping, so the next time the config will still be with the Spreadsheet name instead of with an already exist SpreadsheetId.

@brafdlog WDYT?

PR #127, opened 2020-12-07
Sheets fixes

brafdlog:
- Fix hash column (the current code is fragile because it assumes the which column is the hash column. Will think of a way to make this safer)
- Throw clear error if no sheet with the expected name exists


baruchiro (approved):
OK, no big change, but note that with the Frontend I already implemented, I have an option to create a new Spreadsheet, so of course if you're creating a new one, you will not have even the Spreadsheet.

This lead me to the open issue- #120

I don't care if you merging this, but I think it's not relevent. Unless you think otherwise so that we need to talk.

*Written from [Github CLI](https://github.com/cli/cli/releases) so no auto English fixes*

brafdlog:
> OK, no big change, but note that with the Frontend I already implemented, I have an option to create a new Spreadsheet, so of course if you're creating a new one, you will not have even the Spreadsheet.

Thats true but in any case, the spreadsheet is something the user can change and if deletes/renames the sheet we will fail and I would like to get a clear error so I think it is still relevant.

PR #135, opened 2020-12-15
Show clear progress indication

### LINKED ISSUE #101: Show the progress of the import and export process
Create a ui that conveys clearly the current status of importing/exporting.

This will have two parts:

1. An accounts state in the ui code that gets updated by listening to events.
2. A component showing each account and its current status.

brafdlog:
Closes #101 

progress

- [x] Implement account state that gets updated by listening to events
- [x] Make more room for status indication
- [ ] ~~Move account state to the store~~ (skipping for now)
- [x] Show the company logos on the account cards
- [x] Simplify event types and make them consistent
- [x] Add the option to show the full log of the account, not just the last message
- [x] In case of a general error, show a modal


baruchiro:
> - [ ] Move account state to the store

Can you explain why? Is it just to easily access that data?

Because please note that the store state is not saved in the filesystem, because I think that kind of data shouldn't be persisted between launch.

I think the persisted data is for the specific Vuex module, so maybe all you need is to create another module for your data.

brafdlog:
> > * [ ]  Move account state to the store
> 
> Can you explain why? Is it just to easily access that data?
> 
> Because please note that the store state is not saved in the filesystem, because I think that kind of data shouldn't be persisted between launch.
> 
> I think the persisted data is for the specific Vuex module, so maybe all you need is to create another module for your data.

Its not super important. Just to be consistent. As you said, if I do it I will make it a separate module that is not persisted because it doesn't make sense to persist.

brafdlog:
@baruchiro This is ready for review now

src/accountsMetadata.ts
```diff
@@ -0,0 +1,106 @@
...
+    companyKey: CompanyTypes.visaCal,
+    companyName: 'ויזה כאל',
+    logo: buildLogoUrl('visa-cal.jpeg')
+  },
+  [CompanyTypes.discount]: {
+    companyKey: CompanyTypes.discount,
+    companyName: 'דיסקונט',
+    logo: buildLogoUrl('discount.jpeg')
+  },
+  [OutputVendorName.CSV]: {
```
baruchiro, line 80: I think it should be in a different object. You need to seperate the objects, then you have more control when you're looping on the list.

src/accountsMetadata.ts
```diff
@@ -0,0 +1,106 @@
+import { OutputVendorName } from '@/originalBudgetTrackingApp/configManager/configManager';
+import { CompanyTypes } from 'israeli-bank-scrapers-core';
```
baruchiro, line 2: I'm thinking about the `israeli-bank-scrapers` updates. If we want to derived their list to attach some custom data, I think it will be good if we validate that we calling the `CompanyTypes` from the `israeli-bank-scrapers` only in one place.

Because now, we're calling this also in the `configManager`, and it can create a mismatch.

src/accountsMetadata.ts
```diff
@@ -0,0 +1,106 @@
...
+    logo: buildLogoUrl('max.jpeg',)
+  },
+  [CompanyTypes.amex]: {
+    companyKey: CompanyTypes.amex,
+    companyName: 'American Express',
+    logo: buildLogoUrl('americanExpress.jpeg')
+  },
+  [CompanyTypes.beinleumi]: {
+    companyKey: CompanyTypes.beinleumi,
+    companyName: 'הבינלאומי',
```
baruchiro, line 27: I think we need a font for Hebrew... just note

src/components/app/MainContent.vue
```diff
@@ -36,48 +68,56 @@ const statusToColor = {
...
+    const config = store.getters.Config;
     const scrapingStatus = ref('NOT_STARTED' as keyof typeof statusToColor);
+    const generalError = ref<string>('');
     const inProgress = computed(() => scrapingStatus.value === 'IN_PROGRESS');
     const btnColor = computed(() => statusToColor[scrapingStatus.value]);
-    const entries = ref([] as LogEntry[]);
+    const showAccountCards = computed(() => scrapingStatus.value !== 'NOT_STARTED');
 
-    const eventPublisher = LogsEventEmitter((entry) => entries.value.push(entry));
+    const accountsState = reactive(new AccountsState(config.getActiveImporters, config.getActiveExporters));
```
baruchiro, line 82: I think because only the **initial value** is comes from the **store**, when the store changed, the `accountsState` not updating.

For example, try to run once (to load the component) then add another Importer or enable Exporter, and it will not show.

src/components/app/accountsState.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export class ImporterState extends AccountState {
+
+}
+
+export class ExporterState extends AccountState {
+
+}
+
+export class AccountsState {
```
baruchiro, line 37: Maybe rename to `ScrapingState` to different from the very similar name `AccountState`?

src/components/app/accountsState.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.importers = [];
+    this.exporters = [];
+
+    importersConfig.forEach(({ id, name }) => {
+      this.importers.push(new ImporterState({ id, name }));
+    });
+
+    exportersConfig.forEach(({ id, name }) => {
+      this.exporters.push(new ExporterState({ id, name }));
+    });
```
baruchiro, line 52:
```suggestion
    this.importers = importersConfig.map((importer) => new ImporterState(importer));
    this.exporters = exportersConfig.map((exporter) => new ExporterState(exporter));
```

src/components/app/accountsState.ts
```diff
@@ -0,0 +1,80 @@
+// eslint-disable-next-line max-classes-per-file
```
baruchiro, line 1: Note- We need to note if this file coupled to the UI, or it can define a general way to handle and track events, and if so, it should be in the "backend" side.

src/originalBudgetTrackingApp/eventEmitters/EventEmitter.ts
```diff
@@ -19,53 +20,87 @@ export enum EventNames {
...
-}
+  vendorId?: CompanyTypes | OutputVendorName;
+
+  accountStatus?: AccountStatus;
 
-interface ImporterEndEvent extends ImporterEvent {
- transactions: EnrichedTransaction[]
+  error?: Error;
+
+  accountType?: AccountType;
```
baruchiro, line 40: Ok, it also related to my other comment. I'm not sure how much the Importers and Exporters are similar, and if they should have the same logic.
But lets keep it for now, until we will be smarter.

baruchiro:
Also need a way to return the Config Editor, because you will not see it after your first click on Run.

src/accountsMetadata.ts
```diff
@@ -0,0 +1,106 @@
...
+
+type AccountKey = CompanyTypes | OutputVendorName;
+
+export type AccountMetadata = {
+  companyKey: string;
+  companyName: string;
+  logo: string;
+}
+
+const PUBLIC_STORAGE_PREFIX = 'https://budget-tracking-public.s3.eu-central-1.amazonaws.com/logos';
```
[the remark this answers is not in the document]
brafdlog: No good reason. Will change

src/accountsMetadata.ts
brafdlog: I will export it from the originalBudgetTracking index file so the ui code will not import it directly from the israeli-bank-scrapers library to give us more control

src/accountsMetadata.ts
brafdlog: ok. Lets deal with that when we do some more investing in hebrew

src/components/app/accountsState.ts
brafdlog: I thought about it. But it isn't scraping state because it includes the exporters too.

It actually is AccountsState because AccountState is of a single account and this is basically a container for many of those AccountState objects :) 

WDYT?

src/components/app/accountsState.ts
brafdlog: Good question. I assumed this is ui state (I was planning to move it to the store) and thats why I put it under the ui code. If we see need for such state in the backend code, either will move this there or create another state there with just what the backend needs.

src/components/app/MainContent.vue
brafdlog: Thats true. Solved this by creating a new AccountsState every time we start scraping. This actually makes sense because this basically contains the status of the process so every time we run again the status starts from scratch.

src/components/app/accountsState.ts
baruchiro: Fine, it's not important.

Just because I didn't understand it immediately until I figure that there is class `AccountsState` and `AccountState`.

In VSCode you don't have problems like that.

baruchiro (approved):

PR #138, opened 2020-12-15
Run importers and exporters concurrently

brafdlog:


src/originalBudgetTrackingApp/export/exportTransactions.ts
```diff
@@ -13,8 +13,7 @@ export async function createTransactionsInExternalVendors(
   const executionResult = {};
   const allTransactions = _.flatten(Object.values(companyIdToTransactions));
 
-  for (let j = 0; j < outputVendors.length; j++) {
-    const outputVendor = outputVendors[j];
+  const exportPromises = outputVendors.map(async (outputVendor) => {
```
baruchiro, line 16: Two points just about coding practices, this specific code bothers me from the first time ;-)

I don't like **loop** with **if** inside, I'm preferring `filter` and `map` in this case.

src/originalBudgetTrackingApp/export/exportTransactions.ts
```diff
@@ -31,7 +30,9 @@ export async function createTransactionsInExternalVendors(
         throw e;
       }
     }
-  }
+  });
+
+  await Promise.all(exportPromises);
```
baruchiro, line 35: I wanted to ask you to return the `executionResult` from the promise, instead of updating the `executionResults` from the closure.

But in that case, if one of the promises will fail, you will not get any result, even from the successes promises.

Anyway, have you tried [`Promise.allSettled()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled)? I heard about but never tried...

(This is also just comment on my opinion about nodeJS code)

baruchiro (approved):
Fine, my code would have looked different, but it's a matter of taste.

src/originalBudgetTrackingApp/export/exportTransactions.ts
brafdlog: Thats a good point. Our current behaviour is that if one of them fails they all fail and it would be better to be able to allow all to run and fail only what failed. This will also require the ui to show this correctly.

This is a small and focused pr so I don't want to mix this in

PR #151, opened 2020-12-20
Fixes

### LINKED ISSUE #120: Sheets Exporter: Support new Spreadsheet, new sheet, exists data

brafdlog:
1. Add importers with active: true
2. Prevent the prepare script from overriding the categoryCalculationScript if one exists
3. Implement the create spreadsheet flow (current ui implementation is quick and dirty) Fixes #120 

Part of this change is to reduce the auth scopes we require the user to give us. Before this change we asked for scopes that are considered risky and this means the user gets a special warning and google wants to do a manual review of the app. In this pr, we use a single scope that only gives the app access to spreadsheets that the app has created. Since this scope is not considered risky the user is not warned and the app is approved.


scripts/copyCategoryCalculationScript.js
```diff
@@ -0,0 +1,15 @@
...
+const CATEGORY_CALCULATION_SCRIPT_FILE_NAME = CATEGORY_CALCULATION_SCRIPT_EXAMPLE_FILE_NAME.replace('-example', '');
+
+const CATEGORY_CALCULATION_SCRIPT_EXAMPLE_FILE_PATH = path.join(CATEGORY_CALCULATION_SCRIPT_DIR_PATH, CATEGORY_CALCULATION_SCRIPT_EXAMPLE_FILE_NAME);
+const CATEGORY_CALCULATION_SCRIPT_FILE_PATH = path.join(CATEGORY_CALCULATION_SCRIPT_DIR_PATH, CATEGORY_CALCULATION_SCRIPT_FILE_NAME);
+
+if (!fs.existsSync(CATEGORY_CALCULATION_SCRIPT_FILE_PATH)) {
+  // eslint-disable-next-line no-console
+  console.log('Creating the category calculation script from the example script');
+  fs.copyFileSync(CATEGORY_CALCULATION_SCRIPT_EXAMPLE_FILE_PATH, CATEGORY_CALCULATION_SCRIPT_FILE_PATH);
+}
```
baruchiro, line 15:
```suggestion
const SCRIPT_DIR_PATH = path.join(__dirname, '..', 'src/backend/import');
const SCRIPT_EXAMPLE_FILE_NAME = 'categoryCalculationScript-example.ts';
const SCRIPT_FILE_NAME = SCRIPT_EXAMPLE_FILE_NAME.replace('-example', '');

const SCRIPT_EXAMPLE_FILE_PATH = path.join(SCRIPT_DIR_PATH, SCRIPT_EXAMPLE_FILE_NAME);
const SCRIPT_FILE_PATH = path.join(SCRIPT_DIR_PATH, SCRIPT_FILE_NAME);

if (!fs.existsSync(SCRIPT_FILE_PATH)) {
  // eslint-disable-next-line no-console
  console.log('Creating the category calculation script from the example script');
  fs.copyFileSync(SCRIPT_EXAMPLE_FILE_PATH, SCRIPT_FILE_PATH);
}
```

Lines are too long :-)

I just removed the `CATEGORY_CALCULATION_` to read the code

src/backend/export/outputVendors/googleSheets/googleSheets.ts
```diff
@@ -48,12 +47,34 @@ const createTransactionsInGoogleSheets: ExportTransactionsFunction = async (
...
   await eventPublisher.emit(EventNames.EXPORTER_PROGRESS, new ExporterEvent({
     message, exporterName: googleSheetsOutputVendor.name, allTransactions
   }));
 }
 
+export async function createSpreadsheet(spreadsheetTitle: string, credentials: Auth.Credentials): Promise<string> {
+  const auth = createClient(credentials);
+  const spreadsheetId = await googleSheets.createSpreadsheet(spreadsheetTitle, DEFAULT_SHEET_NAME, auth);
+
+  const columnHeaders = ['תאריך', 'סכום', 'תיאור', 'תיאור נוסף', 'קטגוריה', 'מספר חשבון', 'hash - לא לגעת'];
```
baruchiro, line 71: I think move them up right after `const DEFAULT_SHEET_NAME = '_budget-tracking';`

src/backend/export/outputVendors/googleSheets/googleSheets.ts
```diff
@@ -48,12 +47,34 @@ const createTransactionsInGoogleSheets: ExportTransactionsFunction = async (
   return spreadsheetAppendResult.data;
 };
 
+async function getSheetName(spreadsheetId, oAuthClient) {
+  const sheetNames = await googleSheets.getSheetNames(spreadsheetId, oAuthClient);
+  if (sheetNames.length === 1) {
+    return sheetNames[0];
+  }
```
baruchiro, line 54: I don't like this, I think.

You have a very specific unexpected logic here, that you can't guess, I think.

(I'm trying to figure what bother me)

I think the user will prefer to get an error, in case there is no `_budget-tracking`, instead of rewrite existed data. (And I know, maybe you can't re-write existed data because in the next actions it will throw because there is unexpected data in the sheet. But this is coupling).

BTW, why not creating the sheet in case it does not exist? (You have a code to create the sheet, you just need to export it to a separated function)
Or getting the sheet name from the user?

src/backend/export/outputVendors/googleSheets/googleAuth.ts
```diff
@@ -8,8 +8,7 @@ export const clientId = GOOGLE_CLIENT_ID;
 export const clientSecret = GOOGLE_CLIENT_SECRET;
 export const redirectUri = homepage;
 export const scopes = [
-  'https://www.googleapis.com/auth/drive.metadata.readonly',
-  'https://www.googleapis.com/auth/spreadsheets',
+  'https://www.googleapis.com/auth/drive.file'
```
baruchiro, line 11: You talked about you changed the scopes, and I thought you changed the technique.
Can you explain (here or on Whatsapp) what is the problem, you solved?
Because I thought we need to declare the tightest scopes.

src/ui/components/app/exporters/googleSheets/SpreadsheetExporter.vue
```diff
@@ -89,8 +90,20 @@ export default defineComponent({
...
+      const isNewSheet = exporter.value.options.spreadsheetId.length < 30;
+      if (isNewSheet) {
+        status.value = Status.LOADING;
+        const spreadsheetId = await createSpreadsheet(exporter.value.options.spreadsheetId, exporter.value.options.credentials);
+        exporter.value.options.spreadsheetId = spreadsheetId;
+        status.value = Status.LOGGED_IN;
+      }
+      return dataToReturn.submit();
+    };
+
```
baruchiro, line 103: I did almost the same code (because this is not a complicated one), but I did it in the `SheetsCombobox`. Now I think your code is better, only because I needed to manually call the `events` and I don't like that (although it should be OK if you're creating your own input component)

```js
const createNewSpreadsheet = async () => {
  if (isNewSpreadsheet.value && auth.value) {
    creating.value = true;
    const newSpreadsheet = await createSpreadhseet(props.value!, auth.value).then(({ data }) => data);
    creating.value = false;
    userSpreadsheets.value.push({ id: newSpreadsheet.spreadsheetId, name: newSpreadsheet.properties?.title });
    listeners.input?.(newSpreadsheet.spreadsheetId);
    listeners.change?.(newSpreadsheet.spreadsheetId);
  }
};
```

src/ui/components/app/exporters/googleSheets/SpreadsheetExporter.vue
```diff
@@ -89,8 +90,20 @@ export default defineComponent({
       }
     };
 
+    const submit = async () => {
+      const isNewSheet = exporter.value.options.spreadsheetId.length < 30;
```
baruchiro, line 94: Interesting, but OK.

baruchiro (commented):
Please wait, I didn't finish to review. I want to run locally and I want to compare it to my code.

src/backend/export/outputVendors/googleSheets/googleSheets.ts
brafdlog: I agree this is not optimal.

We can choose to expose the sheet name as a setting if we will see that it is requested.

For now I will just throw an error, thats fine.

brafdlog: This was a change I made during development that no longer makes sense. Will revert this change

src/backend/export/outputVendors/googleSheets/googleAuth.ts
brafdlog: Google classifies different scopes by how sensitive they are. If they are considered sensitive they give the user a warning and require manual approval of the app by google. This scope for example: `...auth/spreadsheets` gives the app read and write access to all of the user's spreadsheets. That is way more than we need and they consider this sensitive.

Instead in this implementation I use only one scope - `...auth/drive.file` and this scope allows the app access to the spreadsheets that it created. This a much less sensitive permission and so it doesn't show the warning and doesn't require google to review and approve the app.

Screen Shot 2020-12-22 at 15 13 07

src/ui/components/app/exporters/googleSheets/SpreadsheetExporter.vue
brafdlog: Thats the hacky part

src/backend/export/outputVendors/googleSheets/googleAuth.ts
baruchiro: Cool

baruchiro (approved):
OK very good!

PR #167, opened 2020-12-23
fix #166 Exporter is ready even it invalid

### LINKED ISSUE #166: Exporter is ready even it invalid
https://github.com/brafdlog/budget-tracking/blob/457c7f82fa8f26f055fb79462ec6e75cc47187af/src/ui/components/app/exporters/exportersCommon.ts#L14

Try to open the `JsonExporter` and clear the textBox. Click outside the textBox and you will see the button is enabled, even the form should be invalid.

The first step is to fix the line above to be `const readyToSave = computed(() => validated.value && changed.value);`, but it still not solving the problem.

Try to use `:disabled="!readyToSave"` instead of `:disabled="!readyToSave.value"`, it helps me before, but now no.

baruchiro:
fix #166

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,16 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
+		"dbaeumer.vscode-eslint",
+		"eamodio.gitlens",
+		"github.vscode-pull-request-github",
+		"octref.vetur"
+	],
+	// List of extensions recommended by VS Code that should not be recommended for users of this workspace.
+	"unwantedRecommendations": [
+		
+	]
+}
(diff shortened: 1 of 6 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,16 @@
...
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
+		"dbaeumer.vscode-eslint",
+		"eamodio.gitlens",
+		"github.vscode-pull-request-github",
+		"octref.vetur"
+	],
```
baruchiro, line 11: Just a funny VSCode feature that gives us the ability to suggest features to developers who cloned the project.

src/ui/components/app/exporters/exportersCommon.ts
```diff
@@ -11,7 +11,7 @@ export function setupExporterConfigForm<T extends OutputVendorName>(exporterName
 
   const validated = ref(true);
   const changed = ref(false);
-  const readyToSave = computed(() => validated && changed);
+  const readyToSave = computed(() => validated.value && changed.value);
```
baruchiro, line 14: This was the bug. Since `validated` and `changed` themself are `Object`s, the `readyToSave` was always `true`.

brafdlog (approved):

PR #168, opened 2020-12-23
TypeScript Misc

baruchiro:


package.json
```diff
@@ -11,15 +11,18 @@
...
     "lint": "eslint --ext .js,.vue,.ts -f ./node_modules/eslint-friendly-formatter src test",
     "lint:fix": "yarn lint --fix",
-    "test": "yarn unit && yarn e2e",
     "typeCheck": "tsc --noEmit",
+
     "unit": "vue-cli-service test:unit",
-    "e2e": "vue-cli-service test:unit --config='test/e2e/jest.e2e.config.js'"
+    "e2e": "vue-cli-service test:unit --config='test/e2e/jest.e2e.config.js'",
+    "test": "yarn unit && yarn e2e"
   },
```
baruchiro, line 26: Just spaces

src/backend/index.ts
```diff
@@ -11,7 +11,6 @@ import outputVendors from './export/outputVendors';
 import * as bankScraper from './import/bankScraper';
 
 export { CompanyTypes } from 'israeli-bank-scrapers-core';
-export { printYnabAccountData } from './setupHelpers';
```
baruchiro, line 14: I don't think we need that here, I moved it to a folder called `src/manual`.

I don't like that script, I want it as CLI :-)

src/logging/sentry.ts
```diff
@@ -0,0 +1,38 @@
...
+      body,
+      ...extra,
+    },
+  });
+};
+
+export default {
+  initializeReporter,
+  userReportProblem
+};
```
baruchiro, line 38: Not really a new file, just converting it from js

src/backend/configManager/encryption/credentials.js
```diff
@@ -1,78 +0,0 @@
-/* eslint-disable no-use-before-define */
```
baruchiro, line 1: File is unused

src/backend/index.ts
brafdlog: My plan is to add a button in the ynab config that will print these for the user to see. If this is a cli, the user will need to install an additional tool. I think it is useful because setting up ynab is complex now

brafdlog (approved):

src/backend/index.ts
baruchiro: OK but since it is still a workaround I think you can get it from `manual/setupHelpers.ts` until we will have a full UI solution for that.

PR #172, opened 2020-12-25
Bug: ENOTDIR, not a directory

### LINKED ISSUE #170: Bug: ENOTDIR, not a directory
This is a bug report from the app interface.
- [ ] Open in this repo (currently opened in the old one)
- [ ] Typo in the title
- [ ] Add OS and App version to the bug report

## Detailes

I installed the app, then I configured my bank accounts.
After that, I configured the JsonExporter to the full path: `C:\Users\BaruchR\source\repos\transactions\data\transactions.json` and the SheetsExporter I configured to create a new spreadsheet (because it has no access to exists spreadsheets).
I didn't change to Config.

Now I pressed on the RUN button and almost immediately I got the error: ENOTDIR.

The problem is also the message is not descriptive, I don't know where it comes from, and there are no logs about that.

image
You can see לאומי don't have an *import error*. Don't know if it points to something.
## Log
```
[2020-12-23 21:50:26.651] [info] Welcome to budget-tracking log
[2020-12-23 21:50:26.661] [info] Version: 0.1.3-beta
[2020-12-23 21:50:38.471] [info] The renderer process got the logger
[2020-12-23 21:57:33.357] [info] Welcome to budget-tracking log
[2020-12-23 21:57:33.376] [info] Version: 0.1.3-beta
[2020-12-23 21:57:45.838] [info] The renderer process got the logger
[2020-12-24 09:30:18.589] [info] Welcome to budget-tracking log
[2020-12-24 09:30:18.605] [info] Version: 0.1.3-beta
[2020-12-24 09:30:22.235] [info] The renderer process got the logger

```

Reference: [download-chromium#download](https://github.com/juliangruber/download-chromium#download)

baruchiro:
Fixes #170

In the first commit- https://github.com/brafdlog/budget-tracking/pull/172/commits/4bcdb0d87af44e086efaff58ac73e0c228f38cfb I just tried to avoid multi download. But I think we still have an issue with the executable location (in dev env it downloaded to the `node_modules` by default, and I think it have an execute permission problem).

In the commit https://github.com/brafdlog/budget-tracking/pull/172/commits/812be09277887d58800de45b6ab6d43452fe293f, I injected the `chromePath` from the point before we starting to scrape each importer.

But then I thought the backend shouldn't depend on `electron`, it should just do its work, so it can get an optional download path (or just use the default like it was before my change), and the UI that know the Electron limitations, should inject the Electron specific path. See https://github.com/brafdlog/budget-tracking/pull/172/commits/97e3e1fe53c021fca879312ef276d86ed600b7ba


brafdlog:
I understand why you want to inject the path. Its a valid solution, but in my opinion it is not worth it (passing the parameter all over). The chrome path is an internal issue of the scraper code.

An alternative is to save a const of the base path in the filesystem and each code that needs to access the file system will work with that const (this applies to configManager as well?). This will centralise this logic to one variable. In the future if we will want a mode without electron we can set this variable as part of the run.

Both solutions are valid. Up to you

brafdlog (approved):
I understand why you want to inject the path. Its a valid solution, but in my opinion it is not worth it (passing the parameter all over). The chrome path is an internal issue of the scraper code.

An alternative is to save a const of the base path in the filesystem and each code that needs to access the file system will work with that const (this applies to configManager as well?). This will centralise this logic to one variable. In the future if we will want a mode without electron we can set this variable as part of the run.

Both solutions are valid. Up to you

baruchiro:
I'm taking your suggestion mainly because the reason this path is not only related just to `chrome`. But now I see it affects only on the high-level, since you still need to get the resolved path to give it to the scraper (you also can use the `download-chromium` again only to resolve it).
Nevermind, the important thing is to not get stuck on these changeable things.

We need something like DependencyInjection, I don't know how implementing this design pattern in JS, if any.

Need to use this path in #173, I will comment there.

PR #174, opened 2020-12-28
Anonymous analytics tracking

### LINKED ISSUE #145: Add analytics tracking to the app
Add basic analytics so we know how many users are using the app and how often

brafdlog:
Closes #145 

Added analytics that are sent to segment. 

We have full control over the data that is sent so we can make sure no personal or sensitive information is sent.

What is being tracked:

1. "Page views" - how many times the app was loaded
2. Scraping events - just the name of the vendor, name of event and if is error or not.

We don't identify the user himself. Instead we use a hash of the machine id as the identifier so we don't event have the actual machine id, just the hash.

### DIFF
--- a/globals.js
+++ b/globals.js
@@ -3,4 +3,5 @@ module.exports = {
   // https://console.developers.google.com/apis/credentials
   GOOGLE_CLIENT_ID: JSON.stringify(process.env.GOOGLE_CLIENT_ID),
   GOOGLE_CLIENT_SECRET: JSON.stringify(process.env.GOOGLE_CLIENT_SECRET),
+  SEGMENT_WRITE_KEY: JSON.stringify(process.env.SEGMENT_WRITE_KEY),
 };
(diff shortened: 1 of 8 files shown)


baruchiro:
> We don't identify the user himself. Instead we use a hash of the machine id as the identifier so we don't event have the actual machine id, just the hash.

Is `node-machine-id` return the hashed machine ID? Because I don't see where you hashing in yourself.

baruchiro (approved):
OK.

My suggestion is to take this event type as "unstable" until we will finish the Beta.
I mean, if we fill we want to send other data, in other structure, it is OK.

brafdlog:
> > We don't identify the user himself. Instead we use a hash of the machine id as the identifier so we don't event have the actual machine id, just the hash.
> 
> Is `node-machine-id` return the hashed machine ID? Because I don't see where you hashing in yourself.

The library hashes it.



> OK.
> 
> My suggestion is to take this event type as "unstable" until we will finish the Beta.
> I mean, if we fill we want to send other data, in other structure, it is OK.

Yes, this will probably change over time

PR #178, opened 2021-01-01
Replace download-chromium with puppeteer-core

baruchiro:
Still have problems with the Windows installed version. I'm getting errors from `download-chromium` when it uses `ncp`. We don't need that package anymore.

Reference: https://github.com/juliangruber/download-chromium/pull/23

Still related to #170


src/ui/components/app/exporters/YnabExporter.vue
```diff
@@ -14,14 +14,14 @@
...
-      :rules="rules.required"
+      :rules="[rules.required]"
       @change="changed = true"
     />
     <v-text-field
       v-model="exporter.options.budgetId"
       label="Budget id"
       outlined
-      :rules="rules.required"
+      :rules="[rules.required]"
```
baruchiro, line 24: Fix bug

src/backend/eventEmitters/EventEmitter.ts
```diff
@@ -26,7 +26,11 @@ export enum AccountType {
 }
 
 export enum AccountStatus {
-  IDLE, PENDING, IN_PROGRESS, DONE, ERROR
+  IDLE = 'IDLE',
+  PENDING = 'PENDING',
+  IN_PROGRESS = 'IN_PROGRESS',
+  DONE = 'DONE',
+  ERROR = 'ERROR'
 }
```
baruchiro, line 34: More readable in the console, so why not

baruchiro:
@brafdlog It is not including another PR now, you can review.

baruchiro:
@brafdlog Note I enabled *Auto-Merge*, I don't know exactly what is the procedure, but it is mean that if you're commenting a suggestion and then approving my PR, it will merge automatically.

This is something to consider. I think in my company we are enabling *Auto Merge* only on already approved PRs, just for auto merging after the validation checks.

brafdlog (approved):

brafdlog:
@baruchiro I am used to approving prs with comments if in general I accept the pr except for minor changes. The benefit of this is that it doesn't require another review which can be a redundant blocker.

So we can either stay with auto merge and only approve when all is well or disable auto merge.

I think we should disable auto merge because waiting for another review on minor changes can take a few days if we are busy and I don't think that is worth it. WDYT?

baruchiro:
Yeah, sure, I totally agree with you. I just wanted to check this feature :-)

Issue #171, opened 2020-12-24
Unexpected behaviour when scraping in parallel and one account failed.


baruchiro:
I wanted to ask you to return the `executionResult` from the promise, instead of updating the `executionResults` from the closure.

But in that case, if one of the promises will fail, you will not get any result, even from the successes promises.

Anyway, have you tried [`Promise.allSettled()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled)? I heard about but never tried...

(This is also just comment on my opinion about nodeJS code)

_Originally posted by @baruchiro in https://github.com/brafdlog/budget-tracking/pull/138#discussion_r545740231_

If one of the scrapers failed, the `Promise.all` will not stop the other scrapers, so on one hand the RUN button will be available and you will feel like the whole process ended, but on the other hand, events continued to fire, so suddenly you will see updates in the UI.

PR #186, opened 2021-01-16
Multi fixes

baruchiro:


src/ui/components/app/importers/DeleteImporterDialog.vue
```diff
@@ -1,41 +1,39 @@
 <template>
```
baruchiro, line 1: I just removed the root `v-row`.

Before:
image

After:
image

brafdlog (approved):

baruchiro:
@brafdlog Still in draft, I need your approval in the other one

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.154.2/containers/javascript-node/.devcontainer/base.Dockerfile
```
baruchiro, line 1: I just tried to run on WSL/Docker, because I have a lot of issues with my computers.

Of course, it problematic because we are trying to launch a UI application.

So I think the Docker/WSL is working (don't remember which one) but with a very slow UI.

src/backend/consts.ts
```diff
@@ -1,7 +1,14 @@
...
+import { mkdirSync } from 'fs';
 import path from 'path';
 
 const App = app || remote.app;
 
+if (process.env.NODE_ENV !== 'production') {
+  const localUserData = path.resolve('userData');
+  mkdirSync(localUserData, { recursive: true });
+  App.setPath('userData', localUserData);
+}
```
baruchiro, line 11: Only for development

src/backend/import/importTransactions.ts
```diff
@@ -19,27 +19,25 @@ type ScrapingConfig = Config['scraping'];
...
-  });
-  await Promise.all(scrapingPromises);
+  const scrapePromises = scrapingConfig.accountsToScrape
+    .filter((accountToScrape) => accountToScrape.active !== false)
+    .map(async (accountToScrape) => ({
+      id: accountToScrape.id,
+      transactions: await fetchTransactions(accountToScrape, startDate, scrapingConfig.showBrowser, eventPublisher, dowloadedChrome)
+    }));
+
+  const promiseResults = await Promise.allSettled(scrapePromises);
```
baruchiro, line 31: The main change is to use `Promise.allSettled(scrapePromises)`, for solving #171.

Instead of using a closure, I prefer to reduce the results from the promises. After trying to export the `map` content into a different function, I found the function is very similar to the `fetchTransactions`, just adds ErrorHandling and emits, so I merged those two functions.

src/backend/import/importTransactions.ts
```diff
@@ -19,27 +19,25 @@ type ScrapingConfig = Config['scraping'];
...
+  const companyIdToTransactions = promiseResults
+    .reduce((idToTrxAcc, scrapeRes) => {
+      if (scrapeRes.status === 'fulfilled') {
```
baruchiro, line 34: If you doing this `if` in a `filter` method, TS will not identify the `scrapeRes` as a type of `FullfilledResult`, and you don't have access to `.value`.

src/ui/components/app/Importers.vue
```diff
@@ -7,29 +7,12 @@
...
-                  :color="iconClass(importer.status.success).color"
-                  dark
-                  v-on="on"
-                >
-                  {{ iconClass(importer.status.success).icon }}
-                </v-icon>
-              </template>
-              <span>{{ importer.status.lastMessage }}</span>
-            </v-tooltip>
-          </template>
```
baruchiro, line 32: Remove unused icon in importers

src/ui/components/app/MainContent.vue
```diff
@@ -79,6 +80,7 @@ export default defineComponent({
     const inProgress = computed(() => scrapingStatus.value === 'IN_PROGRESS');
     const btnColor = computed(() => statusToColor[scrapingStatus.value]);
     const showAccountCards = computed(() => scrapingStatus.value !== 'NOT_STARTED');
+    const enableRun = computed(() => config.getActiveImporters.length && config.getActiveExporters.length);
```
baruchiro, line 83: disable RUN button if no importers/exporters

src/ui/components/shared/log/AccountCardStatusIndicator.vue
```diff
@@ -11,12 +11,30 @@
...
+    <v-card>
+      <v-card-title class="headline red lighten-2">
+        Error
+      </v-card-title>
+      <v-card-text class="general-error-text subtitle-1">
+        {{ errorMessage }}
+      </v-card-text>
+    </v-card>
+  </v-dialog>
+
```
baruchiro, line 37: When clicking on the Error icon, you will see the error message:

Screenshot from 2021-02-02 22-30-47

src/backend/import/importTransactions.ts
baruchiro: So now it considered Green even if some importers failed. WDYT?

Screenshot from 2021-02-02 22-05-39

baruchiro:
@brafdlog Ready for review.

Sorry for this collection, I'm just trying to work with the app and fixing things on the fly.

I'm really suggesting you review commit-by-commit, I think you can see the comments also in the commit view.

Start with: https://github.com/brafdlog/budget-tracking/pull/186/commits/6818f44e2f85ff8a345680c23f2160bde4f27b68

Then click on "next" in the upper right.

brafdlog:
I don't follow the why you need the dockerfile logic. Lets discuss that. 

Other than than, all looks good

baruchiro:
> I don't follow the why you need the dockerfile logic. Lets discuss that.
> 
> Other than than, all looks good. Nice work :)

@brafdlog, about the docker, this is just trying something. Do you know [*VSCode Dev Containers*](https://code.visualstudio.com/docs/remote/containers)?

I just like this feature, and also I have issues in my Windows machine for running this project right now, so I tried to create a Docker for a development workspace. But if you think it confusing, I can leave it in a separate branch.

Except this, any other comments? (I'm not sure you approved, since you already approved early.)

brafdlog (approved):

Issue #103, opened 2020-09-15
Use repo name and URL from package.config


baruchiro:
Instead of using the `budget-tracking` or `https://github.com/brafdlog/budget-tracking` in many places in the log, we need to declare them once and use the variable all over the code.

It can be implemented by Webpack `DefinePlugin` or `import { name } from 'package.json`.

baruchiro:
Please also update with the new name: "Hiuvi".

See #140 and https://github.com/brafdlog/budget-tracking/discussions/139#discussioncomment-343082

baruchiro:
https://github.com/brafdlog/budget-tracking/blob/e9e3159981fa67c7b97f74a013e609f23f61efa8/src/ui/components/App.vue#L53

https://github.com/brafdlog/budget-tracking/blob/e9e3159981fa67c7b97f74a013e609f23f61efa8/src/main.ts#L33

https://github.com/brafdlog/budget-tracking/blob/e9e3159981fa67c7b97f74a013e609f23f61efa8/src/ui/components/App.vue#L10

baruchiro:
Reopen: rename to כספיון
https://github.com/brafdlog/hiuvi/discussions/139#discussioncomment-885000

Issue #140, opened 2020-12-16
Choose a name for the app and update the code


brafdlog:
Choose a name for the app and update the code to that name.

Related discussion #139

baruchiro:
I think the real work is by #103 

@brafdlog I suggest changing the repo name. Don't worry, Github will redirect the old URL to the new one for a time.

baruchiro:
Also, of course, the repo path in the `package.json` should be updated.

MordRot:
And app name in: `globals.js`

baruchiro:
@brafdlog Please change the repo name.
Don't worry, it will redirect the current URL to the new one for some time. (AFAIK)

baruchiro:
Reopen: rename to כספיון
https://github.com/brafdlog/hiuvi/discussions/139#discussioncomment-885000

PR #192, opened 2021-02-16
add global app name and repo #103

MordRot:
Use repo name and URL from package.config #103

### DIFF
--- a/src/ui/components/App.vue
+++ b/src/ui/components/App.vue
@@ -7,7 +7,7 @@
       color="primary"
     >
       <div class="d-flex align-center">
-        <span class="shrink mr-2 headline">Israeli Bank Scrapers Desktop</span>
+        <span class="shrink mr-2 headline"> {{ appName }} </span>
       </div>
 
       <v-spacer />
@@ -50,7 +50,7 @@
       <v-btn
         text
         small
-        @click="() => openExternal('https://github.com/baruchiro/israeli-bank-scrapers-desktop')"
+        @click="() => openExternal({appRepository})"
       >
         <span class="mr-2">Open in Github</span>
         <v-icon>mdi-open-in-new</v-icon>
@@ -71,6 +71,7 @@ import Exporters from '@/ui/components/app/Exporters';
 import { trackPage } from '@/analytics';
 import ReportProblemDialog from './app/ReportProblemDialog';
 import MainContent from './app/MainContent';
+import appProperties from '../../../package.json';
 
 export default {
   name: 'App',
@@ -79,7 +80,9 @@ export default {
   },
   data() {
     return {
-      reportProblemDialog: false
+      reportProblemDialog: false,
+      appRepository: appProperties.repository,
+      appName: APP_NAME
     };
   },
   mounted() {
--- a/src/ui/components/App.vue
+++ b/src/ui/components/App.vue
@@ -50,7 +50,7 @@
       <v-btn
         text
         small
-        @click="() => openExternal({appRepository})"
+        @click="() => openExternal(appRepository)"
       >
         <span class="mr-2">Open in Github</span>
         <v-icon>mdi-open-in-new</v-icon>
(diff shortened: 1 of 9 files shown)


src/ui/components/App.vue
```diff
@@ -71,6 +71,7 @@ import Exporters from '@/ui/components/app/Exporters';
 import { trackPage } from '@/analytics';
 import ReportProblemDialog from './app/ReportProblemDialog';
 import MainContent from './app/MainContent';
+import appProperties from '../../../package.json';
```
baruchiro, line 74:
```suggestion
import { repository } from '../../../package.json';
```

Since you need this object only for the `repository` field, you can extract the `repository` field into variable with the same name

src/ui/components/App.vue
```diff
@@ -50,7 +50,7 @@
       <v-btn
         text
         small
-        @click="() => openExternal('https://github.com/baruchiro/israeli-bank-scrapers-desktop')"
+        @click="() => openExternal({appRepository})"
```
baruchiro, line 53:
```suggestion
        @click="() => openExternal(appRepository)"
```

The `openExternal` is just a function in another place, and it gets a URL as arguemnt, so you don't need the `{ }` (if you use them, it actualy mean that you passed an object with field `appRepository`, instead of only string)

baruchiro:
@MordRot Please see my comments.

Also, if you know about things that should be changed after changing the repository path, I will appreciate if you can list them in #140

baruchiro (approved):
I think there is a problem in ElectronBuilder, not related to you, I will check it later.

vue.config.js
```diff
@@ -27,8 +27,8 @@ module.exports = {
         'israeli-bank-scrapers-core'
       ],
       builderOptions: {
-        productName: 'israeli-bank-scrapers-desktop',
-        appId: 'com.electron.israeli-bank-scrapers-desktop',
+        productName: globals.APP_NAME,
+        appId: 'com.electron.' + globals.APP_NAME,
```
baruchiro, line 31: This is actually not good.

You can see in globals, we not just creating an object with keys and values like this:

```js
const clobals = {
  APP_NAME: "Hiuvi"
}
```

Instead, we using `JSON.stringify`, so the value is a string with quates. Becuase when it replace it in the code, it should put the quates also. It confusing, but at the end you get `appId: 'com.electron."Hiuvi"`

This is why the build failed.

I see some options here:

1. Just set the value here.
2. Export the `APP_NAME` itselfs from the `globals.config.js` file, alongside the clobals (I'm not sure how to do that).
3. Keep the `APP_NAME` value in the `package.json` file, with custom field, see [here](https://stackoverflow.com/a/27232456/839513).

baruchiro: See this line in the build: https://github.com/brafdlog/budget-tracking/pull/192/checks#step:8:92

MordRot: Maybe I'll just use the name from the package.json.
It probably also makes more sense to use the name without uppercase.
The disadvantage is importing anther thing.

L2: `import {name as appName} from './package.json';`

L30: `builderOptions: {`
          `productName: appName,`
          `appId: 'com.electron.' + appName,`

baruchiro: I'm not sure I understand the disadvantage. Do you mean that adding the `import` line is the disadvantage?

I think it is really OK.

vue.config.js
```diff
@@ -1,4 +1,5 @@
 const globals = require('./globals');
+import {name as appName} from './package.json';
```
baruchiro, line 2:
```suggestion
import { name as appName } from './package.json';
```

vue.config.js
```diff
@@ -1,4 +1,5 @@
 const globals = require('./globals');
+import { name as appName } from './package.json';
```
baruchiro, line 2:
```suggestion
const appName = require('./package.json').name;
```

@MordRot It confusing, but you can't use the `import` **syntax** outside a *Module* file.

Just remember that in the config files in the root folder, most of the time you don't have all the language features you have inside the application code.

I'm going to commit it.

PR #193, opened 2021-02-25
Log refactoring

### LINKED ISSUE #141: Define and implement a flow for sending logs to us in case of an error
A user should be able to send us the logs in case he faces an error.

- [ ] Make the logger write to a file
- [ ] Verify there is a global exception handler that logs exceptions
- [ ] Add to the readme a description of how to send logs (either manual or automatic)
- [ ] Make sure we printing things to the log alongside the UI (scraping events, for example)
- [ ] Make sure we are re-assigning the console.log function, or we linting to not use it and use our custom logger instead.

baruchiro:
Close #141

> A user should be able to send us the logs in case he faces an error.
> 
> * [x]  Make the logger write to a file
> * [x]  Verify there is a global exception handler that logs exceptions
> * [x]  Add to the readme a description of how to send logs (either manual or automatic)
> * [ ]  Make sure we printing things to the log alongside the UI (scraping events, for example)
> * [ ]  Make sure we are re-assigning the console.log function, or we linting to not use it and use our custom logger instead.


src/backend/eventEmitters/compositeEventPublisher.ts
```diff
@@ -1,11 +0,0 @@
...
-
-export function buildCompositeEventPublisher(eventPublishers: EventPublisher[]): EventPublisher {
-  const compositeEmitter = new BudgetTrackingEventEmitter();
-
-  compositeEmitter.onAny(((eventName, eventData) => {
-    // @ts-ignore
-    eventPublishers.map((eventPublisher) => eventPublisher.emit(eventName, eventData));
-  }));
-  return compositeEmitter;
-}
```
baruchiro, line 11: This is not a special logic, that should be part of the "eventEmitters" tools. Anyone knows how to write a function like this.

You may remember my conflict with the "EventEmitter", so I removed some codes here, to remove complexity, since all we want from the "backend user" is to give us an object that only implement an interface with `emit`.

src/backend/eventEmitters/consoleEmitter.ts
```diff
@@ -1,11 +0,0 @@
...
-import { BudgetTrackingEventEmitter, EventPublisher } from '@/backend/eventEmitters/EventEmitter';
-
-export function buildConsoleEmitter(): EventPublisher {
-  const consoleEmitter = new BudgetTrackingEventEmitter();
-  consoleEmitter.onAny((eventName, eventData) => {
-    // eslint-disable-next-line no-console
-    console.log(`${eventName}:`, eventData);
-  });
-  return consoleEmitter;
-}
```
baruchiro, line 11: As backend (think about the option to implement CLI in the future, for example), you not sure you will print to the console, you leave it to the "backend user" decision.

baruchiro: But see the new `loggerEmitter`, it is very similar.

src/backend/eventEmitters/loggerEmitter.ts
```diff
@@ -0,0 +1,13 @@
...
+  info: (...params: any[]) => void
+}
+
+export function buildLoggerEmitter(logger: Logger): EventPublisher {
+  const loggerEmitter = new BudgetTrackingEventEmitter();
+  loggerEmitter.onAny((eventName, eventData) => {
+    logger.info(`${eventName}:`, eventData);
+  });
+  return loggerEmitter;
+}
```
baruchiro, line 13: This is not a very smart method, but it is here for a reason.

Here, in the `eventEmitters` folder, in the context of the `eventEmitters`, we know the event types, so we can parse them to string (with special care for each event), and direct them to the right level.

So, continuously, we can use this method to

1. Print specific message depends on the Event.
2. Print to the console or to the file, by level (because the `logger` argument supports both `console.[level]` and `logger.[level]`

src/logging/logger.ts
```diff
@@ -1,29 +1,24 @@
-import { App } from 'electron';
-import logger from 'electron-log';
+import { App } from '@/app-globals';
+import logger, { LogFunctions, LogLevel } from 'electron-log'; // eslint-disable-line no-restricted-imports
```
baruchiro, line 2: If you will try to `import logger from 'electron-log'`, you will get an *ESLint* error, because we want you to use our custom-defined logger.
But here we have to import the logger from `electron-log` to configure it.

src/logging/logger.ts
```diff
@@ -1,29 +1,24 @@
...
-  const onError = (error: Error) => {
-    logger.error(error.message ? error.message : error);
-    if (error.stack) logger.debug(error.stack);
-  };
-  logger.catchErrors({ onError });
+const onError = (error: Error) => {
+  logger.error(error.message || error);
+  if (error.stack) logger.debug(error.stack);
+};
+logger.catchErrors({ onError });
```
baruchiro, line 16: Note, this catching only the `UnhandledRejection`.

At least for the Vue part, we need to open a new issue to catch exceptions during the render process (I even can't say what exactly happened during such case. For in React, for example, the component tree will crash- if you're not using error boundary)

baruchiro: #198

README.md
```diff
@@ -31,6 +31,13 @@ The first time you run the app, you will need to set up the accounts you want to
 Now you can set up the exporters - where the data will be sent to. The CSV exporter is enabled by default.
 If you want to export to YNAB, see instructions at the end of the README.
 
+### Report a problem
+
+We are still in beta, and you may find errors.
+Please use the **REPORT A PROBLEM** button in the app to report to us.
+
+Use this button to find the **logs folder** as well.
+
```
baruchiro, line 40: Please review

src/ui/components/app/ReportProblemDialog.vue
```diff
@@ -60,6 +60,8 @@
             </v-row>
           </v-container>
           <small>*indicates required field</small>
+          <br>
+          <small>You can find the logs here: {{ logsDir }}</small>
```
baruchiro, line 64: image

baruchiro:
@brafdlog Ready for review.

I skipping the two checkboxes. This PR is already big.

PR #204, opened 2021-03-17
Update README.md

brafdlog:
This was an annoying issue....


baruchiro (approved):
Yes, and it hard to find out what exactly we need and which ways we have to do that

PR #222, opened 2021-04-04
Batch dependabot updates

baruchiro:


.github/dependabot.yml
```diff
@@ -1,18 +1,19 @@
...
-- package-ecosystem: npm
-  directory: "/nuxt/"
-  open-pull-requests-limit: 0
-  schedule:
-    interval: "weekly"
+  - package-ecosystem: npm
+    directory: "/nuxt"
+    open-pull-requests-limit: 0
+    schedule:
+      interval: "weekly"
```
baruchiro, line 19: I hope it will stop to suggest upgrades for `nuxt`

.github/dependabot.yml
```diff
@@ -1,18 +1,19 @@
...
-  ignore:
-  - dependency-name: electron
-  - dependency-name: spectron
-  - dependency-name: typescript
-  - dependency-name: sass-loader
-    versions: ["11.x"]
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "monthly"
```
baruchiro, line 6: It create a lot of noises

.github/dependabot.yml
brafdlog: Why would it stop from this change? If you don't want it to udpate nuxt, why not remove this whole chunk?
```suggestion
```

baruchiro: You know, I thought about that. It is OK for me, worst case we can tag the last commit with `nuxt` to easily restore it if we will want to.
I will do that in a different PR, let's merge this PR as is.

brafdlog (approved):

PR #252, opened 2021-06-03
Support hash changes and migrations

### LINKED ISSUE #247: Support hash changes and migrations
Today, after `israeli-bank-scrapers` [fixed the Leumi scraper](https://github.com/eshaham/israeli-bank-scrapers/pull/565), I ran the application again, and I got duplicated transactions in my *Json* and *Google Sheets*.

After checking it I found at least two changes:

1. In my salary transactions, in the `memo`, there is one char changed: ' => `
2. The date of the transaction changed from `2021-01-04T22:00:00.000Z` to `2021-01-04T22:00:00.002Z` (two milliseconds)

For a more robust hash, I guess we should ignore these kinds of differences.

I think we need to implement a function to compare hashes.

----

Now I'm thinking.... Why we have hash? Why are we not calculating it on demand?
I can't remember.

baruchiro:
Fixes #247


src/backend/import/importTransactions.test.ts
```diff
@@ -1,40 +0,0 @@
-import { TransactionStatuses, TransactionTypes } from 'israeli-bank-scrapers-core/lib/transactions';
```
baruchiro, line 1: This file was duplicated with `transactions.test.ts`

src/backend/transactions/transactions.ts
```diff
@@ -3,20 +3,24 @@ import { EnrichedTransaction } from '../commonTypes';
 import { Transaction } from '../import/bankScraper';
 import { compareObjectsByDate } from './dates';
 
-const transactionArrayToObject = (transactions: EnrichedTransaction[]) => transactions.reduce((acc, enrichedTransaction) => {
-  acc[enrichedTransaction.hash] = enrichedTransaction;
+const unifyHash = (hash: string) => hash
+  .replace(/`/g, "'")
+  .replace(/00\dZ/, '000Z');
```
baruchiro, line 8: This is the change. It is part of the hash creation, but also used when comparing hashes to support adding rules.

test/unit/specs/__snapshots__/transactions.test.ts.snap
```diff
@@ -0,0 +1,34 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Transactions Merge Transactions 1`] = `
```
baruchiro, line 3: Snapshot Testing. The test should always generate the exact output. If not, it will show the diff.

It is my preferred way to validate complicated objects.

test/unit/specs/transactions.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  hash: calculateTransactionHash(trx, companyId, accountNumber)
+});
+
+describe('Transactions', () => {
+  test.each([
+    [TRANSACTION_1, TRANSACTION_1_HASH],
+    [TRANSACTION_2, TRANSACTION_2_HASH],
+    [{ ...TRANSACTION_1, processedDate: '2020-06-01T21:00:00.002Z' }, TRANSACTION_1_HASH],
+    [{ ...TRANSACTION_1, date: '2020-03-24T22:00:00.002Z' }, TRANSACTION_1_HASH],
+    [{ ...TRANSACTION_2, description: 'מי גבעתיים בע`מ' }, TRANSACTION_2_HASH]
```
baruchiro, line 50: Changing the `date` to be with `002Z` and the description to be with ` char.

src/backend/transactions/transactions.ts
brafdlog: Manipulating dates with string manipulation is not a very stable way to do it.

Why not use moment, something like `date.startOf('day').format()` with the format of your choice?

baruchiro:
Still not completed, now I have a wired problem with [LRM](https://en.wikipedia.org/wiki/Left-to-right_mark) character. I need to work to ignore it or all the special chars.

src/backend/transactions/transactions.ts
baruchiro: That is because I'm starting with just hash. If I was needed to generate the hash in a better way, for sure I did your suggestion. But here I want also to unify already created hashes from existing Google Sheet, for example, and I can't regenrate the hash based on the sheet data.

PR #253, opened 2021-06-09
Add directory picker/#155

### LINKED ISSUE #155: In CSV and JSON exporter settings use a ui file picker instead of text field to choose where to save the file

baaraak:
Fix #155 
- Add handlers file to register electron new event handlers (and initialize them in `bacgkround.ts`.
- Add custom setup config to CSV (we should use it also in the JSON exporter


baruchiro:
> We still have 2 "bugs"
>
> 1. The user can't enter filePath manually, so basically it's mean that he can't control the generated filename the default is `'transaction.csv' || 'transaction.json'`.
It's ok with you? Also, when the dialog is opened and the user selects a directory path, I get only the path and the file name `'transaction.csv'` is removed.
a. We can add the filename manually something like `filePath = filePath + '/'transaction.csv''`.
b. We can do what I did in the start with the input + icon, then the user can select a path by clicking on the icon and using the dialog and editing it/change the filename.
c. We can let the user choose 1 filename in a central place and all the exporter will use this file name
>
> 2. The json/csv component, I wrote it in here: https://github.com/brafdlog/hiuvi/pull/253/files#diff-e7af9e850b9bf9f5a4b28ef8b5abb00958fde041449aff6497aaeb2e2c40c504R75

_Originally posted by @baaraak in https://github.com/brafdlog/hiuvi/issues/155#issuecomment-857501201_

baruchiro (changes requested):
1. About the concern with the folder/file name, I think you just need to use the [`showSaveDialog`](https://www.electronjs.org/docs/api/dialog#dialogshowsavedialogbrowserwindow-options) instead of `showOpenDialog`. Note the `showSaveDialog` is not actually saving something. It is just the right dialog for selecting a path to a **new** file, instead of selecting a path of **existing** one.
2. Not sure what you mean in the "json/csv component" "bug", maybe I answered you in my review...

Thanks! Good luck!!!!

baruchiro:
@baaraak You can test the production version with the artifacts from [here](https://github.com/brafdlog/hiuvi/actions/runs/922607221).

We will wait for @brafdlog if he wants to take a look, and I will merge it later.

baruchiro (approved):

PR #266, opened 2021-06-17
Fixate version for israeli-bank-scrapers

liady:


package.json
```diff
@@ -40,7 +40,7 @@
     "electron-log": "^4.1.1",
     "emittery": "^0.7.1",
     "googleapis": "^59.0.0",
-    "israeli-bank-scrapers-core": "^1.1.1",
+    "israeli-bank-scrapers-core": "1.1.1",
```
brafdlog, line 43: Welcome🧑‍🚀
Shouldn't this affect the package lock?

baruchiro: I'm not sure.

You know, in my understanding, this is just an easier way to keep track. Theoretically, the packages installed by the `yarn.lock` file, but maybe I'm missing something, please review [this](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install) and maybe we need to use `--frozen-lockfile`.

brafdlog: It does change yarn lock. Please run yarn and submit the updated lock file and then we can merge.

baruchiro (approved):

baruchiro:
@liady Are you allowed to merge?

liady:
@baruchiro currently not

PR #268, opened 2021-06-21
Missing credentials for scraper. CompanyId: hapoalim 

### LINKED ISSUE #257: Missing credentials for scraper. CompanyId: hapoalim
running version 0.2 on mac

when trying to scrape hapoalim I get this error:
```Missing credentials for scraper. CompanyId: hapoalim```

Importing from MAX works.

### PR DESCRIPTION
**Pull request by baruchiro**
fix #257

### DIFF
--- a/src/backend/import/bankScraper.ts
+++ b/src/backend/import/bankScraper.ts
@@ -24,9 +24,6 @@ type EmitProgressEventFunction = (eventCompanyId: string, message: string) => Pr
 export async function scrape({
   companyId, credentials, startDate, showBrowser = false
 }: ScrapeParameters, emitProgressEvent: EmitProgressEventFunction, chromePath: string) {
-  if (!credentials || (!credentials.username && !credentials.num && !credentials.id) || !credentials.password) {
-    throw new Error(`Missing credentials for scraper. CompanyId: ${companyId}`);
-  }
 
   const options = {
     companyId, // mandatory; one of 'hapoalim', 'discount', 'otsarHahayal', 'leumiCard', 'isracard', 'amex'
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/backend/import/bankScraper.ts [hunk 1]
```diff
@@ -24,19 +24,14 @@ type EmitProgressEventFunction = (eventCompanyId: string, message: string) => Pr
 export async function scrape({
   companyId, credentials, startDate, showBrowser = false
 }: ScrapeParameters, emitProgressEvent: EmitProgressEventFunction, chromePath: string) {
-  if (!credentials || (!credentials.username && !credentials.num && !credentials.id) || !credentials.password) {
-    throw new Error(`Missing credentials for scraper. CompanyId: ${companyId}`);
-  }
```
baruchiro, line 29: I removed this. This is a complicated condition and its maintaining worth more than its value.

jossef: suggesting to place the validation logic(s) at the input forms

brafdlog: This validation was added because before we had it, the error in this case was cryptic.

You want to remove it because it fails too often in cases where there are credentials that are valid?

If this is the case, I suggest replacing the exception with a console

brafdlog (approved):

src/backend/import/bankScraper.ts
brafdlog: > suggesting to place the validation logic(s) at the input forms

That's a good idea. Doesn't have to block this PR.

Its actually a good first issue.

baruchiro: > You want to remove it because it fails too often in cases where there are credentials that are valid?
> 
> If this is the case, I suggest replacing the exception with a console

My point is it failed because of a bug in the code, and it is not so important to fail it earlier before the scraping.
I think it is very complicated logic for an almost useless use case. It is complicated because you need to validate the `loginFields` that differ in each scrapper.

> suggesting to place the validation logic(s) at the input forms

And in any way, you can't insert an empty field through the UI.

PR #269, opened 2021-06-22
Rename hiuvi -> Caspion

brafdlog:


src/backend/export/outputVendors/googleSheets/googleSheets.ts
```diff
@@ -10,7 +10,7 @@ import * as googleSheets from './googleSheetsInternalAPI';
 import { appendToSpreadsheet } from './googleSheetsInternalAPI';
 
 const GOOGLE_SHEETS_DATE_FORMAT = 'DD/MM/YYYY';
-const DEFAULT_SHEET_NAME = '_hiuvi';
+const DEFAULT_SHEET_NAME = '_caspion';
```
brafdlog, line 13: This is a breaking change for people who are already using the integration with google sheets, but I think it is worth doing it now and not leaving this old reference in the code.

baruchiro: Yes... but we will need to stop breaking the app someday 😉

.github/workflows/release.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         script: |
           const releases = await github.repos.listReleases({
             owner: 'brafdlog',
-            repo: 'hiuvi'
+            repo: 'caspion'
```
baruchiro, line 39: You need to change the repo in Github

baruchiro (approved):
Just change the repo name in Github and we will see the checks are passed.

PR #270, opened 2021-06-24
send user report

baruchiro:
The user reports didn't showed in Sentry. After trying things it seems we have to enable JS


baruchiro:
@brafdlog It is working: https://sentry.io/organizations/jonathan-wi/discover/caspion:f9a05eb35760486d90c1cb550a37d9e5/

brafdlog (approved):

PR #333, opened 2022-03-24
Add column(s) to Sheets exporter

SeanZoR:
I'm opening this as a draft because I think it still needs thinking/work.

Background: I'm using Caspion with just Google Sheet as exported. Having no "Currency" field is a blocker for me as I can't differentiate when a new "USD" or alike transaction appears.

I've added the changed described in this PR, they do work, but the problem is **it's not backward compatible**.

(Note: The field "Currency" is discussed here, but once we agree on something I can add the rest of fields as appear in the bank scraper or CSV, if needed).

I see a few possible options to make it backward compatible:
1. "Quick and ugly": Instead of adding the field(s) columns in the middle, we can add those in the end (most right column). It's a bit ugly because they will appear after the "...hash..." column. Also, it doesn't promise it won't cause issues for users who are not expecting it. 
2. "Migrating date": Analyzing the structure of the current sheet and manipulating it to extend to new columns. Problem here is that also we're not sure if the sheet itself was manipulated. Also, this is a lot of work with introducing new features on the sheet itself.
3. "Only for new sheets": We can introduce the new fields only for new "sheeters" (🤣). Meaning, if we are creating the sheet for you now, we'll pour in the new fields. This is the most elegant solution IMO, but will require some leftovers in the codebase itself.
4. "Let the users shift the data manually" - we keep the code as it is now, and users will need to shift the code as they see right. meaning a breaking change, but easy to fix.

WDYT?

P.S. Thank you @brafdlog @baruchiro for this amazing code! much appreciated.

### DIFF
--- a/src/backend/export/outputVendors/googleSheets/googleSheets.ts
+++ b/src/backend/export/outputVendors/googleSheets/googleSheets.ts
@@ -11,7 +11,7 @@ import { appendToSpreadsheet } from './googleSheetsInternalAPI';
 
 const GOOGLE_SHEETS_DATE_FORMAT = 'DD/MM/YYYY';
 const DEFAULT_SHEET_NAME = '_caspion';
-const COLUMN_HEADERS = ['תאריך', 'סכום', 'תיאור', 'תיאור נוסף', 'קטגוריה', 'מספר חשבון', 'hash - לא לגעת'];
+const COLUMN_HEADERS = ['תאריך', 'סכום', 'מטבע', 'תיאור', 'תיאור נוסף', 'קטגוריה', 'מספר חשבון', 'hash - לא לגעת'];
 
 const createTransactionsInGoogleSheets: ExportTransactionsFunction = async (
   { transactionsToCreate: transactions, outputVendorsConfig },
@@ -39,6 +39,7 @@ const createTransactionsInGoogleSheets: ExportTransactionsFunction = async (
   const transactionsInSheetsFormat = transactionsToCreate.map((transaction) => [
     moment(transaction.date).format(GOOGLE_SHEETS_DATE_FORMAT),
     transaction.chargedAmount,
+    transaction.originalCurrency,
     transaction.description,
     transaction.memo,
     transaction.category,


brafdlog:
Hi @SeanZoR,
Thanks for bringing this up, these are good points.
I have been thinking about how to handle breaking changes in the structure of the configuration object but haven't yet thought about the spreadsheet structure changing.
Here are my thoughts about the options you suggested, will start with the easier ones :)

(2) Doing an analysis of the structure of the sheet and acting according to that - sounds fragile and not worth the ROI. I don't think we should go that route.
(4) Keeping the code as is - not great. This will break the hashing logic so transactions will be duplicated. The value we gain from the nicer order of columns is not worth IMO such a breaking change.

(3) only for new sheets - this option is interesting. It relates to a thought I was having about versioning. We could make the sheets have versions and each change in the schema will tick up the version in the code. If we do this, we can easily detect that the sheet is not up to date and display a notification to the user with the required changes to comply to the new version (or do the changes for him if he wants). I think long term this is a good option, but given the beta state of the project, I wouldn't go there yet.

(1) Move the column to the end - I think this is the right thing to do now. This gives support for the currency use case at a minimal effort and it doesn't break the sheets backwards because the hashing mechanism will still work so old txns that were created in the old format would not change and just new txns will be created in this format.

@baruchiro  @SeanZoR  WDYT?

brafdlog:
@SeanZoR you can proceed with the first option of moving the column to be the last. @baruchiro is not available this week and we won't wait. If you are happy with this option go ahead, if not, lets discuss

baruchiro:
I agree with @brafdlog, I think we should solve the problems in an easy way, because we don't have enough effort to spend on the best implementation.

Thanks a lot @SeanZoR! Well-defined PR.
Looking forward to seeing more PRs for you, feel free to take care and suggest things, or just contribute for fun :-)
If you need any help, you know where to find us...

(BTW you should set this PR to be *Ready for review* if you want us to approve it..)

SeanZoR:
Hey all, thank you for your notes. I'm on a vacation right now but will
follow through with the changes once I'm back 🔥

SeanZoR:
Hey, OK this looks good. meaning we are going with option `(1) Move the column to the end`.

i had a look on other potential fields and haven't found them useful at this point.

baruchiro (approved):

baruchiro (changes requested):

SeanZoR:
@baruchiro I've made the change now, I believe it's ready. 

I've tested :
A. a migration of it, meaning starting with a sheet that was in `master` branch, and then switching to this branch. Works fine 👍 Looks like this, note the most right column
image

B. Starting a sheet straight from this branch, also looks good and also has the desired col title 
image

About `amount` and `chargedAmount` it's unrelated to currency, but rather just shows the total amount and if you have תשלומים

baruchiro (approved):
OK, approved.
Let me know if you can't merge it and I will merge it for you.

Thank you for a beautiful contribution! Explained and covered edge cases.

What will be your next contribution?

SeanZoR:
Thanks @baruchiro , I don't have `write` access so can't merge, but go ahead and merge for me 👍  

And kudos again for you and @brafdlog on building this project! 👑   It's not easy to dedicate the time to it. 
I don't have any other outstanding issues but if will..

PR #366, opened 2022-08-10
Add missing fields to GoogleSheets exporter

michlus:
The purpose is to add the 6 fields that exist in the csv export file (csv.ts), but are missing from this Google Sheets export. 
Please note that I used to be a programmer, but in C++, not Java or anything after it.  
I have no idea how to test the changes that I propose ... , so this is NOT TESTED.

### DIFF
--- a/src/backend/export/outputVendors/googleSheets/googleSheets.ts
+++ b/src/backend/export/outputVendors/googleSheets/googleSheets.ts
@@ -12,9 +12,9 @@ import { appendToSpreadsheet } from './googleSheetsInternalAPI';
 const GOOGLE_SHEETS_DATE_FORMAT = 'DD/MM/YYYY';
 const DEFAULT_SHEET_NAME = '_caspion';
 const COLUMN_HEADERS = [
-'תאריך', 'סכום', 'תיאור', 'תיאור נוסף',
- 'קטגוריה', 'מספר חשבון', 'hash - לא לגעת',
- 'מטבע','סכום מקורי', 'ת. חיוב', 'מזהה', 'סוג', 'סטטוס'
+  'תאריך', 'סכום', 'תיאור', 'תיאור נוסף',
+  'קטגוריה', 'מספר חשבון', 'hash - לא לגעת',
+  'מטבע', 'סכום מקורי', 'ת. חיוב', 'מזהה', 'סוג', 'סטטוס'
 ];
 
 const createTransactionsInGoogleSheets: ExportTransactionsFunction = async (
--- a/src/backend/export/outputVendors/googleSheets/googleSheets.ts
+++ b/src/backend/export/outputVendors/googleSheets/googleSheets.ts
@@ -12,9 +12,19 @@ import { appendToSpreadsheet } from './googleSheetsInternalAPI';
 const GOOGLE_SHEETS_DATE_FORMAT = 'DD/MM/YYYY';
 const DEFAULT_SHEET_NAME = '_caspion';
 const COLUMN_HEADERS = [
-  'תאריך', 'סכום', 'תיאור', 'תיאור נוסף',
-  'קטגוריה', 'מספר חשבון', 'hash - לא לגעת',
-  'מטבע', 'סכום מקורי', 'ת. חיוב', 'מזהה', 'סוג', 'סטטוס'
+  'תאריך',
+  'סכום',
+  'תיאור',
+  'תיאור נוסף',
+  'קטגוריה',
+  'מספר חשבון',
+  'hash - לא לגעת',
+  'מטבע',
+  'סכום מקורי',
+  'ת. חיוב',
+  'מזהה',
+  'סוג',
+  'סטטוס'
 ];
 
 const createTransactionsInGoogleSheets: ExportTransactionsFunction = async (
--- a/src/backend/export/outputVendors/googleSheets/googleSheets.ts
+++ b/src/backend/export/outputVendors/googleSheets/googleSheets.ts
@@ -11,7 +11,11 @@ import { appendToSpreadsheet } from './googleSheetsInternalAPI';
 
 const GOOGLE_SHEETS_DATE_FORMAT = 'DD/MM/YYYY';
 const DEFAULT_SHEET_NAME = '_caspion';
-const COLUMN_HEADERS = ['תאריך', 'סכום', 'תיאור', 'תיאור נוסף', 'קטגוריה', 'מספר חשבון', 'hash - לא לגעת', 'תאריך חיוב', 'מזהה', 'סטטוס', 'סוג', 'תשלומים', 'סכום מקור', 'מטבע'];
+const COLUMN_HEADERS = [
+'תאריך', 'סכום', 'תיאור', 'תיאור נוסף',
+ 'קטגוריה', 'מספר חשבון', 'hash - לא לגעת',
+ 'מטבע','סכום מקורי', 'ת. חיוב', 'מזהה', 'סוג', 'סטטוס'
+];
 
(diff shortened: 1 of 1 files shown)


src/backend/export/outputVendors/googleSheets/googleSheets.ts
```diff
@@ -11,7 +11,7 @@ import { appendToSpreadsheet } from './googleSheetsInternalAPI';
 
 const GOOGLE_SHEETS_DATE_FORMAT = 'DD/MM/YYYY';
 const DEFAULT_SHEET_NAME = '_caspion';
-const COLUMN_HEADERS = ['תאריך', 'סכום', 'תיאור', 'תיאור נוסף', 'קטגוריה', 'מספר חשבון', 'hash - לא לגעת', 'מטבע'];
+const COLUMN_HEADERS = ['תאריך', 'סכום', 'תיאור', 'תיאור נוסף', 'קטגוריה', 'מספר חשבון', 'hash - לא לגעת', 'תאריך חיוב', 'מזהה', 'סטטוס', 'סוג', 'תשלומים', 'סכום מקור', 'מטבע'];
```
baruchiro, line 14:
```suggestion
const COLUMN_HEADERS = [
  'תאריך', 'סכום', 'תיאור', 'תיאור נוסף',
  'קטגוריה', 'מספר חשבון', 'hash - לא לגעת',
  'תאריך חיוב', 'מזהה', 'סטטוס', 'סוג',
  'תשלומים', 'סכום מקור', 'מטבע'
];
```

You need to break the line because it is too long.

src/backend/export/outputVendors/googleSheets/googleSheets.ts
```diff
@@ -44,6 +44,12 @@ const createTransactionsInGoogleSheets: ExportTransactionsFunction = async (
     transaction.category,
     transaction.accountNumber,
     transaction.hash,
+    moment(transaction.processedDate).format(GOOGLE_SHEETS_DATE_FORMAT),
+    transaction.identifier,
+    transaction.status,
+    transaction.type,
+    transaction.installments,
+    transaction.originalAmount,
     transaction.originalCurrency
```
baruchiro, line 53:
```suggestion
    transaction.hash,
    transaction.originalCurrency,
    moment(transaction.processedDate).format(GOOGLE_SHEETS_DATE_FORMAT),
    transaction.identifier,
    transaction.status,
    transaction.type,
    transaction.installments,
    transaction.originalAmount
```

I think you need to add new fields to the end (after `transaction.originalCurrency`).

Imagine an existing sheet, today the last column is `originalCurrency`. Now you are adding new data, that contains more fields, and the 8th column was `originalCurrency` and now becomes `processedDate`.

baruchiro (changes requested):
Thank you for contributing!!

It is similar to #333 which already merged, so I accept this change.

We just need to think if and how we can notify the user that we have new columns.

michlus:
Hi!
Thanks.
FIrst - I need to let you know that I'm really not familiar with this
development environment; I'm trying to learn on the go, and I'm trying to
be careful not to do any harm.
Finally I succeeded in setting things up, and I succeeded in testing my
code.
Unfortunately I got an error. The columns are correctly created in the
Google Sheet, but I get an error message in red saying:
"Invalid values[3][11]: struct_value { fields { key: "number" value {
number_value: 3.0 } } fields { key: "total" value { number_value: 3.0 } } }"
I have no idea what to do with that and how to debug it. Sorry.
No data at all is written in the sheet (only the columns).

About your comment that the news fields should be at the utmost right, I
agree.  The reason I did it before the original currency column is because
of the comment of @SeanZor: "* Place the Currency field to the most right".
I understood that as a requirement that it must always be at the utmost
right, but now I understand that I misunderstood it.

Can you guide me how to debug/troubleshoot the changes I did? Thanks,
Michel

src/backend/export/outputVendors/googleSheets/googleSheets.ts
michlus: ok. got it

src/backend/export/outputVendors/googleSheets/googleSheets.ts
michlus: no problem.
note that installments created an error. 
i believe it has something to do with formatting. for instance 3 installments out of 4 is 3/4 - there is a problem text vs number. 
anyway, meanwhile in my newer version i've removed the field installements.

baruchiro:
Hi, I don't know how you running your code, but you can put the `debugger` command inside your code to break the code in that line.
You need the DevTools to be open to the `debugger` will work.

Regarding your error, I guess it says the data in line 3 column 11 is not correct. But need to check this.

michlus:
Hi.
In the meantime I solved the problem.
It was related to the installment field. I removed this field.
I got then a build error because of spaces which I fixed.
Do you indeed see my additional version ?
What’s next?
How/when will my changes make it to the release?
Thanks,
Michel
________________________________

baruchiro:
I think I will test it in my computer and then we will create a new release.

michlus:
Perfect. Thank you.
________________________________

baruchiro (approved):
Approved, I didn't test it because downloading build artifacts is almost impossible because the download is so slow that it gets stuck in the middle of it.

@brafdlog Please merge and I will create a new release and will test it.

src/backend/export/outputVendors/googleSheets/googleSheets.ts
```diff
@@ -11,7 +11,11 @@ import { appendToSpreadsheet } from './googleSheetsInternalAPI';
 
 const GOOGLE_SHEETS_DATE_FORMAT = 'DD/MM/YYYY';
 const DEFAULT_SHEET_NAME = '_caspion';
-const COLUMN_HEADERS = ['תאריך', 'סכום', 'תיאור', 'תיאור נוסף', 'קטגוריה', 'מספר חשבון', 'hash - לא לגעת', 'מטבע'];
+const COLUMN_HEADERS = [
```
brafdlog, line 14: It looks like the order of the columns changed. This will mean that the sheets that will be created going forward will not have the correct header order.
Please retain the same order and add the new columns to the end (also making sure they are the same order as the code you added in transactionsToCreate

brafdlog (changes requested):

src/backend/export/outputVendors/googleSheets/googleSheets.ts
```diff
@@ -11,7 +11,11 @@ import { appendToSpreadsheet } from './googleSheetsInternalAPI';
...
+  'תאריך', 'סכום', 'תיאור', 'תיאור נוסף',
+  'קטגוריה', 'מספר חשבון', 'hash - לא לגעת',
+  'מטבע', 'סכום מקורי', 'ת. חיוב', 'מזהה', 'סוג', 'סטטוס'
+];
```
baruchiro, line 18:
```suggestion
const COLUMN_HEADERS = [
  'תאריך',
  'סכום',
  'תיאור',
  'תיאור נוסף',
  'קטגוריה',
  'מספר חשבון',
  'hash - לא לגעת',
  'מטבע',
  'סכום מקורי',
  'ת. חיוב',
  'מזהה',
  'סוג',
  'סטטוס'
];
```
@brafdlog I listed it one-per-line and I think the order is OK. It is a little confusing and I suggest keeping it like that in one-per-line.

@michlus Please accept this suggestion.

michlus:
I see that a build failed. Seems like a fetching packages issue.
I do see however that "https://registry.yarnpkg.com/rxjs/-/rxjs-6.6.7.tgz: " is reachable, at least - without the colon (:)

brafdlog:
Thanks @michlus!

PR #369, opened 2022-09-04
Add link to discord

baruchiro:
image

### DIFF
--- a/globals.js
+++ b/globals.js
@@ -5,5 +5,6 @@ module.exports = {
   GOOGLE_CLIENT_SECRET: JSON.stringify(process.env.GOOGLE_CLIENT_SECRET),
   SEGMENT_WRITE_KEY: JSON.stringify(process.env.SEGMENT_WRITE_KEY),
   SOURCE_COMMIT_SHORT: JSON.stringify(process.env.GITHUB_SHA),
+  DISCORD_CHANNEL: JSON.stringify('https://discord.gg/XWWg7xvJyS'),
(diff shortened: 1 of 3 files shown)


globals.js
```diff
@@ -5,5 +5,6 @@ module.exports = {
   GOOGLE_CLIENT_SECRET: JSON.stringify(process.env.GOOGLE_CLIENT_SECRET),
   SEGMENT_WRITE_KEY: JSON.stringify(process.env.SEGMENT_WRITE_KEY),
   SOURCE_COMMIT_SHORT: JSON.stringify(process.env.GITHUB_SHA),
+  DISCORD_CHANNEL: JSON.stringify('https://discord.gg/XWWg7xvJyS'),
```
brafdlog, line 8: Why Json stringify a hard coded string?

brafdlog (approved):

globals.js
baruchiro: Because Webpack takes the value and put it "as is".

Let's say you want to replace any `GLOBAL_WEBPACK_VAR` to be the **variable name** `x`. In this case, you need to give Webpack the new **name**: `'x'`. You have to give it as `string` because you can't do it in another way.

So, to replace something to be **a string literal**, you need to wrap the original string with `JSON.stringify`:

```javascript
> 'https://discord.gg/XWWg7xvJyS'
'https://discord.gg/XWWg7xvJyS'
> JSON.stringify('https://discord.gg/XWWg7xvJyS')
'"https://discord.gg/XWWg7xvJyS"'
```

PR #372, opened 2022-09-09
Build: use semantic-release for automatic deploy

### LINKED ISSUE #297: Auto Versioning and Releasing
Using [`semantic-release`](https://semantic-release.gitbook.io), we can automatically produce a new version when, for example, pushing to a `release branch.

I think it requires some configuration, see my suggestions here:
https://dev.to/baruchiro/how-to-setup-auto-semantic-release-4fd8

One thing we need to decide is what will trigger the release.
In my article, pushing to the master will trigger the release, but it's not so good, because you can't collect multiple changes for the version, but any PR will create a new version.

baruchiro:
close #297 

- Build only for Linux on PR validation
- Create a GitHub Release and version by `semantic-release`
- Upload the artifacts to the created Release
- Create a CHANGELOG file

See in #297, currently, the implementation is that committing to the `master` branch will trigger a new release (if the commit messages say so).

I debugged this work in my fork by running the action and creating releases.
Still, I guess we will find some bugs after the merge and we will solve them.
For example, I don't know if our current version format (with the `-beta` suffix) will be supported. Maybe we can omit it. We will see.

Sorry for the meaningless commits, this happened when you working on a pipeline configured in your code.


.github/workflows/ci.yml
```diff
@@ -59,3 +59,43 @@ jobs:
...
+      run: |
+        yarn
+        yarn build --publish never
+    - name: Upload artifact
+      uses: actions/upload-artifact@v1.0.0
+      with:
+        # Artifact name
+        name: ${{ matrix.os }}-artifact
+        # Directory containing files to upload
+        path: dist_electron
```
baruchiro, line 100: On PR, it will build the app only on `linux` (before it was built same as for Release, except it was not published as a Github Release).

.github/workflows/semantic-release.yml
```diff
@@ -0,0 +1,65 @@
...
+      new_release_version: ${{ steps.semantic.outputs.new_release_version }}
+      new_release_published: ${{ steps.semantic.outputs.new_release_published }}
+  
+    steps:
+      - uses: actions/checkout@v3
+      - run: yarn
+      - uses: cycjimmy/semantic-release-action@e1fe1fc00a3729593e87efb2f88475de76d64a24
+        id: semantic
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
baruchiro, line 24: It will first create a new GitHub Release, with release notes and a CHANGELOG file.

It will decide on the version number based on commit messages.

.github/workflows/semantic-release.yml
```diff
@@ -0,0 +1,65 @@
...
+  
+  publish:
+    needs: [release]
+    if: needs.release.outputs.new_release_published == 'true'
```
baruchiro, line 28: Then, if a new release was created (based on commit messages) it will build the multi-platform artifacts and will add them to the Release.

release.config.js
```diff
@@ -0,0 +1,41 @@
+const releaseRules = {
+  preset: 'eslint',
+  releaseRules: [
+    { tag: 'Breaking', release: 'major' },
+    { tag: 'Build', release: 'minor' },
+    { tag: 'Chore', release: 'minor' },
+    { tag: 'Fix', release: 'patch' },
+    { tag: 'New', release: 'minor' },
+    { tag: 'Update', release: 'minor' },
+    { tag: 'Upgrade', release: 'minor' }
```
baruchiro, line 10: The commits should be in this format now, in order to publish a new release based on them.

release.config.js
```diff
@@ -0,0 +1,41 @@
...
+        changelogFile: 'CHANGELOG.md'
+      }
+    ],
+    [
+      '@semantic-release/git',
+      {
+        assets: ['package.json', 'CHANGELOG.md'],
+        // eslint-disable-next-line no-template-curly-in-string
+        message: 'chore(release): ${nextRelease.version} [skip ci]\n\n${nextRelease.notes}'
+      }
```
baruchiro, line 38: It will commit the updated CHANGELOG file

vue.config.js
```diff
@@ -28,7 +28,13 @@ module.exports = {
       ],
       builderOptions: {
         productName: appName,
-        appId: 'com.electron.' + appName,
+        appId: `com.electron.${appName}`,
+        publish: [
+          {
+            provider: 'github',
+            releaseType: 'release'
```
baruchiro, line 35: The default is `draft`, but the release is already created by `semantic-release`, so the type is `release`, and it adding the artifact to the existing release.

PR #388, opened 2022-10-21
Update release parallelism and more

brafdlog:


.github/workflows/semantic-release.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     if: needs.release.outputs.new_release_published == 'true'
     
     strategy:
-      max-parallel: 1
+      max-parallel: 3
```
baruchiro, line 31: I set it to `1` because it takes too long and I suspect in parallel it failed more.

baruchiro (approved):
Good work, and I see you committed with SemanticVersioning, so the merge should be OK.

.github/workflows/semantic-release.yml
brafdlog: Its really slow when it is sequential. Theoretically there shouldn't be any relation between them so the parallelism should not affect success rate. Lets see

PR #391, opened 2022-10-29
Show error message in logs

brafdlog:


brafdlog:
Now the scraping error message is visible in the account logs:
Screenshot 2022-10-29 at 12 47 28

brafdlog:
Updated israeli bank scrapers. [Diff](https://github.com/eshaham/israeli-bank-scrapers/compare/v2.2.0...v3.0.0)

PR #397, opened 2022-11-21
New: Show a badge with number of created txns in exporter

brafdlog:
Code is not super elegant, but does the job.

Screenshot 2022-11-21 at 23 40 53


baruchiro (approved):
Good, thank you! See my comments

PR #406, opened 2022-12-01
Validate New Account Details

### LINKED ISSUE #405: New account details can be empty
The user can add empty accounts without setting any details or with missing fields.

ShiranAbir:
Fixes #405.

### DIFF
--- a/ui-react/src/accountMetadata.tsx
+++ b/ui-react/src/accountMetadata.tsx
@@ -36,28 +36,26 @@ const accountMetadata: Record<CompanyTypes | OutputVendorName, AccountMetadata> 
   };
 });
 
-const PASSWORD_FIELD = 'password';
-
 export const IMPORTERS_LOGIN_FIELDS = {
-  [CompanyTypes.hapoalim]: ['userCode', PASSWORD_FIELD],
-  [CompanyTypes.hapoalimBeOnline]: ['userCode', PASSWORD_FIELD],
-  [CompanyTypes.leumi]: ['username', PASSWORD_FIELD],
-  [CompanyTypes.mizrahi]: ['username', PASSWORD_FIELD],
-  [CompanyTypes.discount]: ['id', PASSWORD_FIELD, 'num'],
-  [CompanyTypes.otsarHahayal]: ['username', PASSWORD_FIELD],
-  [CompanyTypes.leumiCard]: ['username', PASSWORD_FIELD],
-  [CompanyTypes.max]: ['username', PASSWORD_FIELD],
-  [CompanyTypes.visaCal]: ['username', PASSWORD_FIELD],
-  [CompanyTypes.isracard]: ['id', 'card6Digits', PASSWORD_FIELD],
-  [CompanyTypes.amex]: ['id', 'card6Digits', PASSWORD_FIELD],
-  [CompanyTypes.union]: ['username', PASSWORD_FIELD],
-  [CompanyTypes.beinleumi]: ['username', PASSWORD_FIELD],
-  [CompanyTypes.massad]: ['username', PASSWORD_FIELD]
+  [CompanyTypes.hapoalim]: ['userCode', 'password'],
+  [CompanyTypes.hapoalimBeOnline]: ['userCode', 'password'],
+  [CompanyTypes.leumi]: ['username', 'password'],
+  [CompanyTypes.mizrahi]: ['username', 'password'],
+  [CompanyTypes.discount]: ['id', 'password', 'num'],
+  [CompanyTypes.otsarHahayal]: ['username', 'password'],
+  [CompanyTypes.leumiCard]: ['username', 'password'],
+  [CompanyTypes.max]: ['username', 'password'],
+  [CompanyTypes.visaCal]: ['username', 'password'],
+  [CompanyTypes.isracard]: ['id', 'card6Digits', 'password'],
+  [CompanyTypes.amex]: ['id', 'card6Digits', 'password'],
+  [CompanyTypes.union]: ['username', 'password'],
+  [CompanyTypes.beinleumi]: ['username', 'password'],
+  [CompanyTypes.massad]: ['username', 'password']
 };
 
 export const LOGIN_FIELD_DISPLAY_NAMES = {
   'userCode': 'קוד משתמש',
-  [PASSWORD_FIELD]: 'סיסמא',
+  'password': 'סיסמא',
   username: 'שם משתמש',
   id: 'מספר זהות',
   num: 'קוד מזהה',
--- a/ui-react/src/accountMetadata.tsx
+++ b/ui-react/src/accountMetadata.tsx
@@ -63,6 +63,16 @@ export const LOGIN_FIELD_DISPLAY_NAMES = {
 
 };
 
+export const LOGIN_FIELD_MIN_LENGTH = {
+  'userCode': 3,
+  'password': 4,
+  username: 3,
+  id: 9,
+  num: 4,
+  card6Digits: 6,
+
+};
+
 export const importers: Account[] = [];
 
 for (const importerName in CompanyTypes) {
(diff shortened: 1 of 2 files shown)


ShiranAbir:
I've made a new list in `accountMetadata.tsx` that specifies the minimum length for each field.
So this PR checks more than just empty fields :)

ShiranAbir:
I noticed a warning in the console when entering account details:
```
index.js:1 Warning: A component is changing an uncontrolled input to be controlled.
This is likely caused by the value changing from undefined to a defined value, which should not happen.
Decide between using a controlled or uncontrolled input element for the lifetime of the component.
More info: https://reactjs.org/link/controlled-components
```

Read more about this here: https://stackoverflow.com/a/37427596

It became clear that this is related to the `loginFields` list being empty when the controller is initialized.

When form input is changed, the `onLoginFieldChanged` function is called.
The function saves the input in the `loginFields` list. For example:
`loginFields.username='my_username'`

But, `loginFields.username` was never initialized (take note `loginFields` is an empty list=`{}`): https://github.com/brafdlog/caspion/blob/80ce4917cde17fb3eb5e7009f0ce27d56112d39f/ui-react/src/components/accounts/EditImporter.tsx#L18

This is why when ReactDOM checks `loginFields.username` it's undefined and thus considered uncontrolled.
Later, when the user enters card details, the `loginFields.username` suddenly has a value (because `onLoginFieldChanged` set it with `onChange={}`) -> it becomes controlled.

My fix initializes the `loginFields` with each field type, which solves the warning in the console and maintains the form's functionality.

baruchiro:
Nice! I hope I will review it soon :-)

ui-react/src/components/accounts/EditImporter.tsx
```diff
@@ -15,14 +15,23 @@ export default function EditImporter({
   handleDelete,
   importer
 }: EditImporterProps) {
-    const [loginFields, setLoginFields] = useState<Record<string, string>>(importer.loginFields || {});
+    const [loginFields, setLoginFields] = useState<Record<string, string>>(importer.loginFields || {username: '', password: '', userCode: '', id: '', num: '', card6Digits: ''});
```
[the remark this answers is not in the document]
baruchiro: OK, I see your comment in the PR.

Are you saying this default object is not related to your other changes with the field validation?

In any way:
1. I don't like this object initializing inside a `useState`, it is too long. My common solution is to define a `const enptyLoginfFileds` on the head of the code, and use it inside the `useState`.
1. I afraid that with this approach, for each importer it will save the whole object, but with only the relevant fields not empty. Don't know if we want it. (And now, if it makes your solution more complicate, consider solving it in another PR)

ui-react/src/components/accounts/EditImporter.tsx
```diff
@@ -15,14 +15,23 @@ export default function EditImporter({
...
+            event.stopPropagation();
+        } else {
+            await handleSave({
+                ...importer,
+                active,
+                loginFields
+            });
+        }        
+        
+        setValidated(true);
```
baruchiro, line 34: In my optionion, using a Form is unnecessary in React.

Instead, you can just validate the Form on `onSaveClicked` and change the `validated` state.
Also you can enable/disable the Save button by the `validated` state.

This will hekp you avoiding taking care of `preventDefault` and `stopPropagation`.

baruchiro (commented):
Good work!

Thank you for joining us!

ui-react/src/components/accounts/EditImporter.tsx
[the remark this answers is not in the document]
ShiranAbir: Will open a new PR regarding this.

ui-react/src/accountMetadata.tsx
```diff
@@ -36,35 +36,43 @@ const accountMetadata: Record<CompanyTypes | OutputVendorName, AccountMetadata>
...
 export const LOGIN_FIELD_DISPLAY_NAMES = {
   'userCode': 'קוד משתמש',
-  [PASSWORD_FIELD]: 'סיסמא',
+  'password': 'סיסמא',
   username: 'שם משתמש',
   id: 'מספר זהות',
   num: 'קוד מזהה',
   card6Digits: '6 ספרות של הכרטיס',
 
 };
```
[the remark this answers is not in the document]
ShiranAbir: After syncing my fork, branch and merging the mainstream changes, I ran the linter using:
`yarn linter`.
Strangely, the only 3 errors I saw were `no-console` errors on `ynab.ts`.
Is the linter configured correctly?

ShiranAbir:
@baruchiro See the changes in accordance to the comments.

brafdlog (approved):
Nice work.
Thanks Shiran

ui-react/src/components/accounts/EditImporter.tsx
```diff
@@ -17,18 +17,37 @@ export default function EditImporter({
...
+        }
+        
+        for (let fieldKey in loginFields) {
+            if (loginFieldName === fieldKey) continue
+
+            if (!checkFieldValidity(fieldKey, loginFields[fieldKey])) {
+                setValidated(false);
+                return;
+            }
+        }
```
baruchiro, line 49: This code is a little strange for me. Here is what I understand:

1. First, you're trying to validate the current field, and if it is OK, you're validating the others (by skipping the current).  
 I'm not sure what the idea is, why not validating all the fields in one loop?
2. You're right, this validation should happend every time a field is changed. But in React, we can do it with `useEffect`, and implementing this way is better, because it will depend the validation in the state change, and not in the input event.

baruchiro (commented):
One more thing :-)

I assume you're looking for feedbacks, so I wrote another one.

Thanks a lot!

ui-react/src/accountMetadata.tsx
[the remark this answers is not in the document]
ShiranAbir: Sorry meant `yarn lint` (late hours!).
And I ran it before I fixed it, that's why I asked.

ui-react/src/components/accounts/EditImporter.tsx
avivnakar: @baruchiro 
Sorry for interrupting, but I am not sure this is the right use case for `useEffect`. 
have you seen this video? really changed how i look at this hook
https://www.youtube.com/watch?v=HPoC-k7Rxwo

ShiranAbir: Thanks for the video!
@baruchiro Already implemented the `useEffect` but didn't push it yet.
Let me know what you'd like after seeing the video.

Regarding your first question, I wasn't validating all the fields in one loop because the changed field didn't have the updated value (it had the previous version- e.g. you type `c` after `ab` but it only sees `ab`). Probably because `setLoginFields` didn't finish setting it and there was some race condition.

baruchiro: @avivnakar Hi! Nice to see you're participating in our reviews :-)

Usually, I wouldn't say I like videos to learn something new, but I will try to watch for you two.

baruchiro: OK, I sew it and I accept. I don't think it is critical here, and sometimes the readability of the code is important, but let's use **EventHandler** instead of **useEffect** here.

But @ShiranAbir, I still want you to refactor the code to just do one check on the object, for example: `setValidated(Object.entries(loginFields).some(([key, value]) => !checkFieldValidity(key, value)))`

ShiranAbir: @baruchiro Done. Please approve and merge.

baruchiro (approved):
I'm not familiar with `useStateCallback`, and I can't see its documentation.

Approved.

avivnakar (approved):
Good job @ShiranAbir 👑
Here is my approval, just in case it's needed for the protocol

baruchiro:
@avivnakar THANK YOU!

Your approval is not enough for us ;-)

baruchiro:
@ShiranAbir I think you create a regression bug here.

When I want to disable an importer temporarily, the **Save** button is disabled when I toggle the "enable/disable" toggle.

Look:

PR #408, opened 2022-12-05
New: Topbar and Report Problem in the new UI

ErezBiren:
implement open GitHub and open discord buttons + top nav style


ui-react/src/components/topBar/LogsCanvas.tsx
```diff
@@ -0,0 +1,25 @@
...
+      onHide={handleClose}
+      placement="bottom"
+      style={{ zIndex: 9999 }}
+    >
+      <Offcanvas.Header closeButton>
+        <Offcanvas.Title>לוגים:</Offcanvas.Title>
+      </Offcanvas.Header>
+      <Offcanvas.Body>
+        [2022-12-07 12:34:27.432] [info] Welcome to caspion log
+      </Offcanvas.Body>
```
baruchiro, line 22: You missed fetching the real log here, right?

ui-react/src/components/topBar/ReportProblemModal.tsx
```diff
@@ -0,0 +1,248 @@
...
+};
+
+type ReportProblemModalProps = {
+  show: boolean;
+  handleCloseModal: () => {};
+};
+
+function ReportProblemModal({
+  show,
+  handleCloseModal,
```
baruchiro, line 21: You can call it just `onClose`, you don't need to add the `modal` since it will be used always with the `modal` context anyway.

ui-react/src/components/topBar/ReportProblemModal.tsx
```diff
@@ -0,0 +1,248 @@
...
+      if (!form.email || !isValidEmail(form.email))
+        newErrors.email = "שדה מייל לא חוקי";
+    }
+
+    return newErrors;
+  };
+
+  function isValidEmail(email) {
+    return /\S+@\S+\.\S+/.test(email);
+  }
```
baruchiro, line 53: This function need to be outside the component, since it is just a helper function.

ui-react/src/components/topBar/TopBar.tsx
```diff
@@ -1,16 +1,54 @@
...
+    openExternal(repository);
+  };
+
+  const handleDiscord = () => {
+    openExternal(discordchannel);
+  };
+
+  const handleReportProblem = () => {
+    setShowReportModal(true);
+  };
```
baruchiro, line 31: If it is just a function calling to another function, I think you can write the real function just inside the JSX and save us from reading another step just to understand what is append:

`<NavButton onClick={() => setShowReportModal(true)} text="דיווח על בעיה" />`

baruchiro (commented):
Please add screenshots to show what you did.

I commented few points, I need to check it more later.

ui-react/src/components/topBar/LogsCanvas.tsx
ErezBiren: I wanted to consult with you about this,
As I understand, currently the Logger is missing in the ui-react src folder
so we need to add it and start using it there. (correct me if I'm wrong )

and if so, maybe it should be on another PR ?

ErezBiren:
image

new Report Problem dialog

baruchiro:
@avivnakar Here, Erez is completing the migration from Vue to React, by adding some important links and report problem dialog.

He is implementing it by the [Figma design](https://www.figma.com/file/MujqYxhwpWIlweMm5yBiex/Jonathan?node-id=0%3A1), and using the old Vue interface as a reference where we already implemented those items.

ui-react/src/components/topBar/LogsCanvas.tsx
baruchiro: The *Vue* code is using the `logger.ts` file, I don't see a reason to not use it in *React*.

Maybe the challenge is using the same code in a different project, now when the backend is separated from the `react-ui`. @brafdlog What do you suggest for this case?

ui-react/src/components/topBar/NavButton.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { Button } from 'react-bootstrap';
+import navEllipseIcon from "../../assets/navEllipse.svg"
+
+function NavButton({onClick, text}) {
+  return (
+    <div>
+     <img src={navEllipseIcon} alt='navEllipse' style={{cursor: "pointer"}}/>
+      <Button variant="light" onClick={onClick} style={{backgroundColor:"transparent", borderWidth:0}}>{text}</Button> 
+    </div>
```
avivnakar, line 10: I believe that the proper solution would be:
```suggestion
      <Button variant="light" onClick={onClick} style={{backgroundColor:"transparent", borderWidth:0,cursor: "pointer", backgroundImage:navEllipseIcon}}>{text}</Button> 
```
This way, the button and the icon are clickable and the whole thing is more semantic and accessible. 

An img tag should not be used unless the image itself is semantically an element which we would like to provide a blind person description for (alt="three dogs playing with each other") in order to make the page contents make sense for someone who can't see it.

 An icon is not something you would want to describe for a blind person, that's why background will be a better option.

Keep in mind i haven't looked at the figma so this may need some size adjustment for the button to make it fit it's own background image and such

ui-react/src/components/topBar/ReportProblemModal.tsx
```diff
@@ -0,0 +1,247 @@
...
+    title: "",
+    email: "",
+    details: "",
+  });
+
+  const [errors, setErrors] = useState({});
+  const [showLogs, setShowLogs] = useState(false);
+
+  const setField = (field: string, value: string) => {
+    setForm({ ...form, [field]: value });
```
avivnakar, line 34: This solution may cause race condition.
The proper solution would be using callback that will receive the current state and return the next state.

```suggestion
    setForm(prevForm=>({ ...prevForm, [field]: value }));
```
Another valid solution is `useReducer`

ui-react/src/components/topBar/TopBar.tsx
```diff
@@ -1,16 +1,50 @@
...
-      </Container>
-    </NavBar>
+    <>
+      <NavBar
+        className={styles.topNavBar}
+        style={{ borderBottom: "1px solid #C4C4C4" }}
+      >
+        <Container fluid>
+          <NavBar.Brand>
+            <Image
```
avivnakar, line 24: The logo should probably be a div with background instead of an image. For the same reasons as the button + icon

avivnakar (commented):
Here are my thoughts. Hope it is clear my intentions are never to hurt and always to improve. Thank you for your time, hope you will find my comments valuable

ui-react/src/components/Body.tsx
```diff
@@ -61,7 +61,7 @@ const Body = ({ scrape }: BodyProps) => {
 
   return (
     <div>
-      <Container className={styles.container}>
+      <Container className={styles.container} >
```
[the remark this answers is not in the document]
ErezBiren: @avivnakar  - sure thing, thanks for your enlightening inputs!

ui-react/src/utils/validations.ts
```diff
@@ -0,0 +1,4 @@
+
+export function isValidEmail(email: string) {
+    return /\S+@\S+\.\S+/.test(email);
```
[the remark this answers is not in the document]
baruchiro: Do you mean this?
`(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:(2(5[0-5]|[0-4][0-9])|1[0-9][0-9]|[1-9]?[0-9]))\.){3}(?:(2(5[0-5]|[0-4][0-9])|1[0-9][0-9]|[1-9]?[0-9])|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])`


OmgSchockGIF

ui-react/src/components/topBar/LogsCanvas.tsx
```diff
@@ -0,0 +1,25 @@
...
+  handleClose?: () => {};
+};
+
+export default function LogsCanvas({ show, handleClose }: LogsCanvasProps) {
+  return (
+    <Offcanvas
+      show={show}
+      onHide={handleClose}
+      placement="bottom"
+      style={{ zIndex: 9999 }}
```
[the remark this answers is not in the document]
baruchiro: Oh, good!

Simple but essential!

But, Is there no built-in dialog component in the library?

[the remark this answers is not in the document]
ErezBiren: I'm using the _react-bootstrap_ built-in Modal, but also a OffCanvas for the logs list.
And the latter should be on top of the former, which forces me to deal with z-indices.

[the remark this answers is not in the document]
ErezBiren: so I created zIndexesManager.ts which manages the z-indexes layers.
Not sure if I did it right + had to add " //@ts-ignore" since I got some typescript problem there.

ui-react/src/utils/zIndexesManager.ts
```diff
@@ -0,0 +1,17 @@
+
+export let zIndexes:any;
```
avivnakar, line 2:
```suggestion
export let zIndexes:Record<ZIndexLayers, number>
```
This way, anyone who will use this object will have amazing DX with auto completion.

avivnakar (commented):
Here is my Typescript suggestion for the z index management

ui-react/src/utils/zIndexesManager.ts
```diff
@@ -0,0 +1,17 @@
...
+const makeZIndexes = (layers: string[]) =>
+  layers.reduce((prevValue, layerName, index) => {
+     //@ts-ignore
+    prevValue[layerName] = (index + 1) * 10000;
+    return prevValue;
+  }, {});
+
+export const getZIndexes = () => {
+  if (!zIndexes) {
+    zIndexes = makeZIndexes(["modal", "offcanvas"]);
```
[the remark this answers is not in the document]
ErezBiren: It caused some errors,
why not just use:

type ZIndexLayers = "modal" | "offcanvas";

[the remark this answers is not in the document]
ErezBiren: it caused me an error
why not just use:
`type ZIndexLayers = "modal" | "offcanvas";`
?

ui-react/src/components/topBar/ReportProblemModal.tsx
```diff
@@ -0,0 +1,244 @@
...
+    if (Object.keys(formErrors).length > 0) {
+      setErrors(formErrors);
+      return;
+    }
+
+    // const eventId = Sentry.userReportProblem(
+    //   form.title,
+    //   form.details,
+    //   form.attachedLogs ?? "",
+    //   form.email
```
baruchiro, line 113: But I think this is needed, it is not related to the *logs* problem.

Sometimes users clicks on it and I'm getting their report on Sentry.

ErezBiren: Currently, the Sentry.ts file is sitting in the backend just like the logger.ts
and both are using some parameters from the backend env file.
so we should discuss how to use both of them in the new React project, and where the parameters should be held in that case.

But I think it should be in another PR.

baruchiro: Yes, Ok, but it is not good that you commented only the Sentry send function, because now the user will press on **Send** and it will do nothing.

Better is to comment our the button or replace it to open the Discord.

baruchiro (approved):
Very good work!! Thanks a lot!!!
You changed a lot of code and resolved many comments on your first PR, I appreciate that!

Thanks @avivnakar for your review! See you in the next PR :-)

Please fix the [last comment](https://github.com/brafdlog/caspion/pull/408#discussion_r1054268769) and we will merge it!

ui-react/src/components/topBar/ReportProblemModal.tsx
ErezBiren: Sure, I missed this one.
Fixed.

ErezBiren:
@baruchiro 
Thanks your for your warm words and the opportunity to contribute to your project!!!

ui-react/src/components/topBar/ReportProblemModal.tsx
```diff
@@ -0,0 +1,255 @@
...
+      setErrors(formErrors);
+      return;
+    }
+
+    const url = createGithubIssueLink(
+      form.title ?? '',
+      form.details ?? '',
+      form.attachedLogs ?? ''
+    );
+    console.info(`Open bug report url with title: ${form.title}`);
```
baruchiro, line 71:
```suggestion
```

Let's delete it

PR #417, opened 2022-12-11
Chore: use electron-google-oauth2 from @baruchiro

baruchiro:
The original @getstation/electron-google-oauth2 is not support Electron v14:
https://github.com/getstation/electron-google-oauth2/issues/42
And the author is not responding:
https://github.com/getstation/electron-google-oauth2/pull/43

I forked this project and published a new version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,6 @@
   },
   "dependencies": {
     "@electron/remote": "^2.0.8",
-    "@getstation/electron-google-oauth2": "2.1.0",
     "@sentry/electron": "^2.5.0",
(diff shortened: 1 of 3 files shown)


baruchiro:
@avivnakar Here I'm replacing an external library.

This library doing two simple things:
1. Launch a Google Authentication request.
2. Catch the response with a local server.

I read the code before using that library. And now, the library blocks us from upgrading the Electron version, and the author is no longer responding, so I forked it and deployed it as another package, with my fix.

brafdlog (approved):

Issue #420, opened 2022-12-17
On mac montrey when trying to enter new account the application fails


dj-dn:
Hi team,

I am using mac with the pre-built releases (tried 1.6 and 1.5) and when trying to add account (tried bank ha-poalim and bein-leumi)
<img width="254" alt="image" src="https://user-images.githubusercontent.com/104728153/208246332-748c8254-d641-43de-a9c0-182f75954e16.png">
The application cashes.
before crash:
<img width="1153" alt="image" src="https://user-images.githubusercontent.com/104728153/208246404-16ff23e6-3315-459a-b65c-3bef4db93082.png">
when pressing any letter in the account text box the app just turns white:
<img width="1178" alt="image" src="https://user-images.githubusercontent.com/104728153/208246456-f88472fd-1877-408e-b75b-4ea8a8555d29.png">

brafdlog:
Hi @dj-dn,
Please open the developer tools console before doing this and tell us what error you see.
To open the developer tools press: 
Command + Option + I

There you will have a tab called console where you should see the error

dj-dn:
<img width="1147" alt="image" src="https://user-images.githubusercontent.com/104728153/208256562-09f47772-f861-48d6-913f-420f2625ccd3.png">

dj-dn:
Is there any WA in this version (some local file to edit?)
Do you think it will not happen if I run it on Windows VM?

brafdlog:
I see the issue, will create a branch with a fix soon

dj-dn:
@brafdlog will you release a version with the fix or you recommend be to build from source?

brafdlog:
Working on it: https://github.com/brafdlog/caspion/actions/runs/3721745288

PR #421, opened 2022-12-17
Fix: Re add missing import

brafdlog:
Fix issue #420


brafdlog:
@baruchiro this is a regression, a needed import was removed. Not sure how it passed lint in this state

baruchiro:
@brafdlog I asking the same question :-)

PR #433, opened 2023-01-03
New: make the scraping timeout configurable

ErezBiren:
image

### DIFF
--- a/src/backend/commonTypes.ts
+++ b/src/backend/commonTypes.ts
@@ -17,6 +17,7 @@ export interface Config {
     accountsToScrape: AccountToScrapeConfig[];
     chromiumPath?: string;
     maxConcurrency?: number;
+    timeout:number;
   },
   useReactUI?: boolean
 }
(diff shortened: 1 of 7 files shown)


src/backend/commonTypes.ts
```diff
@@ -17,6 +17,7 @@ export interface Config {
     accountsToScrape: AccountToScrapeConfig[];
     chromiumPath?: string;
     maxConcurrency?: number;
+    timeout:number;
```
brafdlog, line 20:
```suggestion
    timeout: number;
```

brafdlog (approved):
Looks good.
Just add spaces between the colon (:) and the type definition.
There are multiple cases where the space is missing

src/backend/commonTypes.ts
ErezBiren: Thanks,
Strange that the eslint did not take care of it for me as it does on the client side.
Fixed.

PR #434, opened 2023-01-04
New: Implement check for updates

ErezBiren:
image

### DIFF
--- a/ui-react/src/components/CheckForUpdates.tsx
+++ b/ui-react/src/components/CheckForUpdates.tsx
@@ -1,0 +1,88 @@
+import React, { useState, useEffect, useContext } from 'react';
+import { Button, ProgressBar, Spinner } from 'react-bootstrap';
+import {
+  checkForUpdate, getAppInfo, openExternal, downloadUpdate
+} from '../eventsBridge';
+import { StoreContext } from '../Store';
+import { styles } from './CheckForUpdates.module.css';
+
+const UPDATE_STATES = {
+  INIT: 'INIT',
+  LOADING: 'LOADING',
+  ERROR: 'ERROR',
+  NEW_VERSION_AVAILABLE: 'NEW_VERSION_AVAILABLE',
+  NO_NEW_VERSION: 'NO_NEW_VERSION',
+  READY_TO_INSTALL: 'READY_TO_INSTALL',
+};
+
+type UpdateInfo = {
+  version: string
+}
+
+function CheckForUpdates() {
+
+  const [updateState, setUpdateState] = useState(UPDATE_STATES.INIT);
+  const [updateInfo, setUpdateInfo] = useState<UpdateInfo>();
+  const store = useContext(StoreContext);
+
+  const checkForUpdates = async () => {
+    setUpdateState(UPDATE_STATES.LOADING);
+    try {
+      const info = await checkForUpdate();
+      setUpdateInfo(info);
+      setUpdateState(info ? UPDATE_STATES.NEW_VERSION_AVAILABLE : UPDATE_STATES.NO_NEW_VERSION);
+    } catch (error) {
+      setUpdateState(UPDATE_STATES.ERROR);
+    }
+  };
+
+  const downloadNewVersion = async () => {
+    setUpdateState(UPDATE_STATES.LOADING);
+    try {
+      await downloadUpdate();
+      setUpdateState(UPDATE_STATES.READY_TO_INSTALL);
+    } catch (error) {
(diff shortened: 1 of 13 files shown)


ErezBiren:
Since I updated the store with a new field called `appInfo` that hasn't been before.
So, I also refactored some of the components to use this new `appInfo` object.

ui-react/src/components/CheckForUpdates.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const store = useContext(StoreContext);
+
+  const checkForUpdates = async () => {
+    setUpdateState(UPDATE_STATES.LOADING);
+    try {
+      const info = await checkForUpdate();
+      setUpdateInfo(info);
+      setUpdateState(info ? UPDATE_STATES.NEW_VERSION_AVAILABLE : UPDATE_STATES.NO_NEW_VERSION);
+    } catch (error) {
+      setUpdateState(UPDATE_STATES.ERROR);
```
brafdlog, line 35: Log the error?
Swallowing errors makes it hard to understand the root cause of issues

ui-react/src/components/CheckForUpdates.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+  };
+
+  const downloadNewVersion = async () => {
+    setUpdateState(UPDATE_STATES.LOADING);
+    try {
+      await downloadUpdate();
+      setUpdateState(UPDATE_STATES.READY_TO_INSTALL);
+    } catch (error) {
+      setUpdateState(UPDATE_STATES.ERROR);
```
brafdlog, line 45: Log

ui-react/src/components/CheckForUpdates.tsx
```diff
@@ -0,0 +1,86 @@
...
+        בדיקת עדכונים
+      </Button>
+    )}
+    {updateState === UPDATE_STATES.NO_NEW_VERSION && (
+      <div>אין עדכונים זמינים</div>
+    )}
+     {updateState === UPDATE_STATES.NEW_VERSION_AVAILABLE && (
+        <>
+          <Button variant="dark" size="sm" className='mb-2' onClick={downloadNewVersion}>
+            Download version {updateInfo?.version}
```
brafdlog, line 72: Why here and compare in English?

brafdlog (approved):
Code looks good. See comments.
The ui design is problematic. The most used button of the app is the הפעל button and we are making the area around it too overloaded.
Its fine for this step but this one of the first things we should solve with the designer

ui-react/src/components/CheckForUpdates.tsx
ErezBiren: Do you mean to log it to the sentry?
Just `console.log(error);`  causing an eslint error:
[](https://eslint.org/docs/rules/no-console)

ErezBiren:
Totally agree about the UI.
This task made me bring a new designer to the project,
Hopefully, she will help us with these things.

ErezBiren:
image

ui-react/src/components/CheckForUpdates.tsx
brafdlog: I think we should disable this eslint rule.
De facto, we need to be able to see the root cause when there is an issue, and since this is not a fancy product, but a real tool we give to people I see no harm in logging the errors to the console. WDYT?

brafdlog:
@ErezBiren looking good, we are just missing the logging part and then I can merge

ErezBiren:
@brafdlog I updated the package.json file with the no-console rule. 
I'm not sure why it still fails.

brafdlog:
This looks like a temporary issue with the registry that is not related to your changes.
Will retry the build
image

ui-react/src/components/CheckForUpdates.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+  };
+
+  const downloadNewVersion = async () => {
+    setUpdateState(UPDATE_STATES.LOADING);
+    try {
+      await downloadUpdate();
+      setUpdateState(UPDATE_STATES.READY_TO_INSTALL);
+    } catch (error) {
+      console.log(error);
```
brafdlog, line 46:
```suggestion
      console.error(error);
```

ErezBiren:
@brafdlog Are you sure it's not the console thing?
image

baruchiro:
@ErezBiren I still don't understand why are you using `console.log` and not `logger` from `src/logging/logger.ts`

PR #438, opened 2023-01-09
Chore: upgrade MacOS build to include Python

baruchiro:
It's solving the MacOS build.


brafdlog (commented):
How did this fix the fact that there is no python?

brafdlog (approved):

PR #445, opened 2023-01-13
New: button for opening exporter destination

ErezBiren:


brafdlog:
Did you run prettier or something? there are a ton of changes that are just styling

ui-react/src/Store.tsx
```diff
@@ -43,8 +45,44 @@ export default class Store {
     });
   }
 
+  setOpenResultsFunction = () => {
```
brafdlog, line 48: This will work, but is suboptimal because this code will keep growing for each new exporter, and you will need to remember for each new exporter to add code here to handle it.
A more elegant way to do this is to add to the exporter interface an openResults section that will contain:
enum ResultType of Browser or FILE
and a URI string that will be either the file path or the website URL.
Then openExporterResults can be a generic function that does:
```
if (exporter.resultType === Browser) {
openBrowser(exporter.uri)
} else {
openFile(exporter.uri)
}
```

This way, the next exporter we add, will be forced to define the config and then will just work.

brafdlog: This will also mean you don't need to do this initial setOpenResultsFunction.
You can just make it openResults(exporter)

ui-react/src/Store.tsx
```diff
@@ -43,8 +45,44 @@ export default class Store {
...
+
+    if (this.config === undefined || !this.config.outputVendors) return;
```
baruchiro, line 50:
```suggestion
    if (!this.config?.outputVendors) return;
```

brafdlog (changes requested):
I tried it and it works very nicely :)
Please see comments

ui-react/src/Store.tsx
brafdlog: The only problem I now see is the dynamic strings.
Since our exporters are defined in config, we don't have a place to put a function field

brafdlog: Another way to achieve this is to use Typescript Record<OutputVendorName> to force an implementation for each exporter
https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type

ErezBiren: Thanks!
Fixed.

brafdlog (approved):

brafdlog:
Thanks @ErezBiren !

ErezBiren:
@brafdlog 
It's my pleasure!
Thanks for the mentoring!

PR #455, opened 2023-01-31
Fix: Scraping indication not shown correctly

### LINKED ISSUE #454: Scraping indication stopped working after enum change

brafdlog:
closes #454


baruchiro (approved):
I see this `enum` exists in both *backend* and *UI*.

Did you try to use the same file in them both?

I know @ErezBiren tried and failed.

brafdlog:
Yes, we need to fix that, but I don't have time for that now
Wanted to release a fix so the app experience won't be broken

PR #457, opened 2023-02-01
Fix: App not installed after "auto update"

### LINKED ISSUE #456: App not installed after "auto update"
I'm updating the app from inside the "Check for updates" button.

After the update is downloaded, I'm clicking on "exit and install", the installation starts, but after it finish I can't find the App anymore.

Windows 11.

baruchiro:
Fix #456


baruchiro:
It seems that nothing is special in this PR.

PR #480, opened 2023-05-29
Fix: Ensure consistent dates for G-Sheets export

berzniz:
This is a fix for https://github.com/brafdlog/caspion/issues/476 ("google spreadsheet dates inconsistent cell format")

The fix is to  use the format of `YYYY-MM-DD` which has no ambiguity and Google Sheets will automatically detect it as a date field.

I did not see any unit-testing in this project so added no tests. If I am missing the tests, I will add.

### DIFF
--- a/src/backend/export/outputVendors/googleSheets/googleSheets.ts
+++ b/src/backend/export/outputVendors/googleSheets/googleSheets.ts
@@ -9,7 +9,7 @@ import { createClient } from './googleAuth';
 import * as googleSheets from './googleSheetsInternalAPI';
 import { appendToSpreadsheet } from './googleSheetsInternalAPI';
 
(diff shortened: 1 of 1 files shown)


baruchiro (approved):
Thanks Tal!!!

So simple... possibly I didn't read the issue carefully and I thought it is so much more difficult (after handling annoying problems with dates with different timezones 😩)

We do have tests, for example, this file `src/backend/export/outputVendors/ynab/ynab.test.ts`. But you're right, the coverage is low, and specific for `googleSheets.ts` there are no tests, and the function combined API calls and data formatting.

You're welcome to refactor the file for testing, but I will not force you to do that 🙂

TheOfficeBowGIF

berzniz:
I see the test suite now... missed it somehow. I can add tests, let me know if that's a prerequisite as there are no tests for Google Sheets exporter today.

PR #494, opened 2023-07-23
Fix: remove another special character to the hash

baruchiro:
Also, don't use yarn 3

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,5 +132,6 @@
   },
   "resolutions": {
     "electron-builder": "^23.0.0"
-  }
+  },
+  "packageManager": "yarn@1.22.19"
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -132,5 +132,6 @@
   },
   "resolutions": {
     "electron-builder": "^23.0.0"
-  }
+  },
+  "packageManager": "yarn@1.22.19"
```
brafdlog, line 136: Why lock this version of yarn?

brafdlog (approved):

package.json
baruchiro: I don't know 😱😱😱
I tried `yarn` version 3 in another project, but it was worst, and I tried to force using `yarn classic` but it returned back to version 3 every time, until I declared it here!

baruchiro:
Tested and it worked

Issue #464, opened 2023-04-02
Upgrade NodeJS version


baruchiro:
I think we are using the latest `v14` version, as you can see here:
https://github.com/brafdlog/caspion/actions/runs/4061787331/workflow#L42

I will try to upgrade the node version anyway.

_Originally posted by @baruchiro in https://github.com/brafdlog/caspion/issues/463#issuecomment-1493247083_

baruchiro:
We are on Node20. Closed

PR #497, opened 2023-09-01
Deps: add prettier config & adapt eslint config

Yedidyar:
what is the motivation:

https://www.youtube.com/watch?v=Cd-gBxzcsdA

if this or accepted we should add format to git hooks

add run yarn format on all project


baruchiro:
Thank you!

Even before it is accepted, I'm expecting you to test it is not conflicted with ESLint.

Did you run it and run `yarn lint` afterwards?

Also, I can't see how it helps. No one will change the whole source code.

Instead, I'm asking you to add it with `lint-staged` to incrementally format every file that is touched by someone.

https://bscstudent.netlify.app/ci-checklist/

baruchiro (approved):
Very good work!

I will test it locally and then I will merge it.

What will be your next task? Contact me

Yedidyar:
> Very good work!
> 
> I will test it locally and then I will merge it.
> 
> What will be your next task? Contact me

I see that the ci dont work because of the node version what should we do?

baruchiro:
@Yedidyar we have a task for upgrading the nodeJS version: https://github.com/brafdlog/caspion/issues/464

But it is not a quick task.

You should check the `lint-staged` and find its version that matches the nodeJS v14.

Yedidyar:
> @Yedidyar we have a task for upgrading the nodeJS version: #464
> 
> But it is not a quick task.
> 
> You should check the `lint-staged` and find its version that matches the nodeJS v14.

[cf4e936](https://github.com/brafdlog/caspion/pull/497/commits/cf4e9365c756ffda64ed2091c3a5fa91a6208129)

package.json
```diff
@@ -51,6 +49,7 @@
     "googleapis": "^87.0.0",
     "israeli-bank-scrapers-core": "3.10.3",
     "keytar": "^7.4.0",
+    "lint-staged": "^13.3.0",
```
baruchiro, line 52: Version `13.3.0` is actually require node 14, as you can see [here](https://github.com/okonet/lint-staged/blob/v13.3.0/package.json#L17), and we need to take one version down.
```suggestion
    "lint-staged": "~13.2.3",
```

baruchiro (changes requested):
There is still a mismatch between `prettier` and `ESLint`.
When I'm running `yarn prettier --write .` and then `yarn lint`, I see multiple warnings like this:

> ⚠  http://eslint.org/docs/rules/object-curly-newline  Expected a line break before this closing brace src\Store.test.tsx:3:56

I'm navigating to this file (`src/Store.test.tsx:3:56`), formatting it with `prettier`, and it shows me an `ESLint` worning. I'm fixing the `ESLint` warning, and formatting it again, and the warning cames back.

Yedidyar:
@baruchiro [26c4866](https://github.com/brafdlog/caspion/pull/497/commits/26c4866a902047b7e0454ad23a65eae5ece92b65) see this commit a think it the right approach

Yedidyar:
> Very good work!
> 
> I will test it locally and then I will merge it.
> 
> What will be your next task? Contact me

I think the next step is to move to using Vite.
what is the specific current version of node?

baruchiro (approved):

PR #519, opened 2023-12-12
Fix: disable category enrichment (fixes #518)

### LINKED ISSUE #518: Add an option to not run the category enrichment
Currently  the category enrichment is always running and since it's not really maintained it overwrites any category coming from the scraper

I think that this feature should be opt in. especially since it requires running from source to change that script

brafdlog:
Fix: #518


brafdlog:
Comment out the category enrichment for now as it is not fully implemented #518

baruchiro (approved):

PR #520, opened 2023-12-13
Chore: add react-hooks lint and typo

baruchiro:
- node v16.19
- fix Scaper typo from israeli-bank-scrapers
- add `eslint-plugin-react-hooks`
- add launch debug configurations


baruchiro:
There is a circular dependency here

PR #529, opened 2024-01-11
Fix: YNAB certificate error on new UI

### LINKED ISSUE #481: YNAB certificate error on new UI
When running Caspion from the new UI there is an error in the YNAB export. No errors occur when running from the old UI.
Error message:
```
starting

request to https://api.youneedabudget.com/v1/budgets/XXXXXX/categories failed, reason: self signed certificate in certificate chain
```

Running on Mac

baruchiro:
Fixes #481

The problem reported on Mac, and reproduced only there, with the React UI.

Based on [this](https://stackoverflow.com/a/63924528/839513) example and others, as we know this API is allowed, we can approve the certificate.

I wanted to verify the certification with more fields, but I can't debug it in my Windows.

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -33,9 +33,9 @@ async function createWindow() {
     width: 1152,
     webPreferences: {
       contextIsolation: false,
-      nodeIntegration: (process.env.ELECTRON_NODE_INTEGRATION as unknown) as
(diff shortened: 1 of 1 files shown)


baruchiro:
The user said that the problem was reproduced with this version.

I think we should wait for better logs to analyze this problem: https://github.com/brafdlog/caspion/issues/428.

PR #533, opened 2024-01-14
try upgrade vue

baruchiro:


baruchiro:
Closing: it's better to remove Vue instead of upgrading it.

PR #574, opened 2024-07-14
Create trivy.yml for vulnerably scan 

sk3pp3r:


brafdlog:
Why is there trivy and trivy2?

.github/workflows/trivy2.yml
```diff
@@ -0,0 +1,48 @@
...
+          docker build -t docker.io/my-organization/my-app:${{ github.sha }} .
+
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@7b7aa264d83dc58691451798b4d117d53d21edfe
+        with:
+          image-ref: 'docker.io/my-organization/my-app:${{ github.sha }}'
+          format: 'template'
+          template: '@/contrib/sarif.tpl'
+          output: 'trivy-results.sarif'
+          severity: 'CRITICAL,HIGH'
```
baruchiro, line 43: Citing code from the internet is legit, but only if you read and sign every line.

Here, you copied a code to scan a Docker image, but we are not using Docker in our project.

baruchiro (changes requested):

PR #585, opened 2024-09-14
Fix: YNAB budget when there is only one budget id

shaiu:
Fix: fix (fixes https://github.com/brafdlog/caspion/issues/462)


baruchiro:
Please manage the clone of this too: https://github.com/brafdlog/caspion/pull/586

@brafdlog please review.

PR #579, opened 2024-08-20
Breaking: major infrastructure upgrade

baruchiro:
Part of #578

PR #586, opened 2024-09-15
Fix YNAB budget when there is only one budget id

baruchiro:
Copy of #585 but for #579

### DIFF
--- a/ui-react/src/components/exporters/EditYnabExporter.tsx
+++ b/ui-react/src/components/exporters/EditYnabExporter.tsx
@@ -2,9 +2,7 @@ import React, { useContext, useEffect, useState } from 'react';
 import { toJS } from 'mobx';
 import Spinner from 'react-bootstrap/Spinner';
 import { observer } from 'mobx-react-lite';
-import {
-  Button, Card, Form, Image
-} from 'react-bootstrap';
+import { Button, Card, Form, Image } from 'react-bootstrap';
 import styles from './EditYnabExporter.module.css';
 import { YnabConfig } from '../../types';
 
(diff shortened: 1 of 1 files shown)


brafdlog:
Hey
Thanks for the PR!
Regarding the formatting, when there are formatting changes like this, please do them in a separate commit, as it makes it hard to see what are the actual changes you want vs just formatting.
Also, why are you changing the formatting?

brafdlog:
I made a related fix here: https://github.com/brafdlog/caspion/commit/b6471362d9cfd91c3e8b515547c55dd7ecd98cad
Not sure why I don't see it in the code

baruchiro:
@brafdlog I made a mistake in this PR. I created a PR from one branch into two target branches: `master` and `vite-electron-builder`. Now we need to merge the `vite-electron-builder` into the PR branch.

And this is why you can't see your change, because there is a conflict.

But your code is there. If it solves the problem, please close this PR.

brafdlog:
@shaiu  if you take a build from [brafdlog:vite-electron-builder](https://github.com/brafdlog/caspion/tree/vite-electron-builder) so you still experience the bug?

brafdlog:
> @brafdlog I made a mistake in this PR. I created a PR from one branch into two target branches: `master` and `vite-electron-builder`. Now we need to merge the `vite-electron-builder` into the PR branch.
> 
> And this is why you can't see your change, because there is a conflict.
> 
> But your code is there. If it solves the problem, please close this PR.

Just saw this.
It should solve the bug, @shaiu  if it doesn't feel free to reopen

PR #587, opened 2024-09-15
vite electron builder tests

shaiu:


baruchiro (commented):
Good fix, but there are many fixes to do
image

image

shaiu:
@baruchiro my bad. It was supposed to be opened as a draft. Still working on it

PR #600, opened 2024-10-07
Fix: bump chromium revision to 1364960

shaiu:
This fixes issue: https://github.com/brafdlog/caspion/issues/599


baruchiro (approved):
Thank you!

PR #601, opened 2024-10-09
Fix: logging and empty config

baruchiro:
This pull request refactors the logging statements in the codebase to use the logger instead of console.log. This improves the consistency and maintainability of the logging system.

The following commits are included in this pull request:

- license package.json

- enable logger

- Refactor logging statements to use logger instead of console.log

### DIFF
--- a/packages/main/src/backend/configManager/configManager.ts
+++ b/packages/main/src/backend/configManager/configManager.ts
@@ -3,6 +3,7 @@ import { type Config } from '@/backend/commonTypes';
 import { decrypt, encrypt } from '@/backend/configManager/encryption/crypto';
 import { existsSync, promises as fs } from 'fs';
 import configExample from './defaultConfig';
+import logger from '/@/logging/logger';
 
 export async function getConfig(
   configPath: string = configFilePath,
@@ -11,13 +12,13 @@ export async function getConfig(
   if (configFromFile) {
     const decrypted = (await decrypt(configFromFile)) as string;
     if (!decrypted) {
-      console.log('No config file found, returning default config');
+      logger.log('No config file found, returning default config');
       return configExample;
     }
     try {
       return JSON.parse(decrypted);
     } catch (e) {
-      console.error('Failed to parse config file, returning default config', e);
+      logger.error('Failed to parse config file, returning default config', e);
     }
   }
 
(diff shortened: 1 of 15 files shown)


packages/main/src/backend/configManager/configManager.ts
```diff
@@ -11,13 +12,18 @@ export async function getConfig(
...
       return configExample;
     }
     try {
-      return JSON.parse(decrypted);
+      const config = JSON.parse(decrypted);
+      if (Object.keys(config).length === 0) {
+        logger.log('Empty config file found, returning default config');
+        return configExample;
+      }
+      return config;
```
baruchiro, line 24: This is a fix for my case, when the config somehow became `{}`. (Locally on dev env)

packages/main/src/logging/logger.ts
```diff
@@ -1,13 +1,22 @@
...
 import { EOL } from 'os';
 import path from 'path';
 
+log.initialize();
+
+// This will transport the logs to the renderer process (DevTools) in production too
+log.transports.ipc.level = log.transports.file.level;
+
+Object.assign(console, log.functions);
+const logger = log.scope('main');
```
baruchiro, line 18: - `initialize`: Get all logs from the renderer into the main logger (the renderer can't write to file)
- `transport.ipc` defaults to `false` for production. Since our users learned to debug the problem with the DevTools, I think it will help to send the main logs into the renderer console.
- `Object.assign(console, log.functions)` to catch also the `console.log`a

packages/main/src/logging/logger.ts
```diff
@@ -18,19 +27,20 @@ const onError: MainErrorHandlerOptions['onError'] = ({ error }) => {
   logger.error(error.message || error);
   if (error.stack) logger.debug(error.stack);
 };
-logger.errorHandler.startCatching({ onError });
-logger.catchErrors({ onError });
```
baruchiro, line 22: deprecated

packages/main/src/logging/logger.ts
```diff
@@ -1,13 +1,22 @@
...
+  MainErrorHandlerOptions,
+} from 'electron-log';
+import log from 'electron-log/main';
 import fs from 'fs';
 import { EOL } from 'os';
 import path from 'path';
 
+log.initialize();
+
+// This will transport the logs to the renderer process (DevTools) in production too
```
shaiu, line 14: Please remove comment

PR #604, opened 2024-10-11
Fix: blank screen after 2.0.5

shaiu:
This fixes the issue in 2.0.5 when loading the app there is a blank screen (no importers and exporters)

Added a default config to ConfigStore constructor

### DIFF
--- a/.nano-staged.js
+++ b/.nano-staged.js
@@ -8,9 +8,7 @@ export default {
    * @param {string[]} filenames
    * @return {string[]}
    */
-  '{package-lock.json,packages/**/{*.ts,*.tsx,tsconfig.json}}': ({
-    filenames,
-  }) => {
+  '{package-lock.json,packages/**/{*.ts,*.tsx,tsconfig.json}}': ({ filenames }) => {
     // if dependencies was changed run type checking for all packages
     if (filenames.some((f) => f.endsWith('package-lock.json'))) {
       return ['yarn typecheck'];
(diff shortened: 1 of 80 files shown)


baruchiro:
I think it will be fixed by #601, if not, I will merge yours.

shaiu:
> I think it will be fixed by #601, if not, I will merge yours.

How exactly? I think I introduced a major bug in 2.0.5 that erases users config..
So this should be merged and also we should remove 2.0.5 from Releases

baruchiro:
> > I think it will be fixed by #601, if not, I will merge yours.
> 
> How exactly? I think I introduced a major bug in 2.0.5 that erases users config.. So this should be merged and also we should remove 2.0.5 from Releases

Do you think it is a bug introduced in 2.0.5?
If so, we didn't find it, we just covered it, and it may return. You and my solutions both fall back to the default config (in my fix I'm doing it right after reading the config, if it is an empty object...). If the bug is still there, users may be reset to the default config without warning.

Anyway, I pushed my fix, and if the bug is still there, we will see it happen and find it.

baruchiro:
@shaiu why did you open it again?

I think it can be closed.

shaiu:
@baruchiro 
I believe that this is the bug:
config: Config = {} as Config;

baruchiro:
Oh, I see it now.

But I prefer you don't create a default config here. We already have the default config on the 'main'/'backend' side, and the `saveConfigIntoFile` will not save it if it is `undefined`. Then, it will try to load it from the backend, and if it is empty or does not exist on the filesystem, the backend will return the default config.

PR #613, opened 2024-11-01
New: Support automatic scraping periodically

brafdlog:
הוספתי אפשרות בהגדרות להגיד שרוצים שירוץ אוטומטית כל כמה שעות
זה נראה ככה:
Screenshot 2024-11-01 at 22 57 09

כשלוחצים על הפעל זה מאתחל את הטיימר של ההרצה האוטומטית.

### DIFF
--- a/packages/main/src/backend/index.ts
+++ b/packages/main/src/backend/index.ts
@@ -43,12 +43,14 @@ export async function scrapeAndUpdateOutputVendors(config: Config, optionalEvent
 
   const startDate = moment().subtract(config.scraping.numDaysBack, 'days').startOf('day').toDate();
 
-  const nextAutomaticScrapeDate : Date | null = config.scraping.periodicScrapingIntervalHours ? moment().add(config.scraping.periodicScrapingIntervalHours, 'hours').toDate() : null;
+  const nextAutomaticScrapeDate: Date | null = config.scraping.periodicScrapingIntervalHours
+    ? moment().add(config.scraping.periodicScrapingIntervalHours, 'hours').toDate()
+    : null;
 
-  await eventPublisher.emit(EventNames.IMPORT_PROCESS_START, new Events.ImportStartEvent(
-    `Starting to scrape from ${startDate} to today`,
-    nextAutomaticScrapeDate,
-  ));
+  await eventPublisher.emit(
+    EventNames.IMPORT_PROCESS_START,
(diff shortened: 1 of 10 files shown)


packages/preload/src/index.ts
```diff
@@ -2,4 +2,10 @@
  * @module preload
  */
 
+import { stopPeriodicScraping } from './eventsBridge';
+
 export * from './eventsBridge';
+
+// Clear the interval that scrapes periodically
+await stopPeriodicScraping();
```
baruchiro, line 10: I'm not sure about this, wrote you on Whatsapp

brafdlog: Removed this

brafdlog:
שיניתי את המימוש שיראה את השעה של הריצה ההבאה.
לא ממשק יפה במיוחד, אבל יעשה את העבודה
Screenshot 2024-11-09 at 14 28 02

packages/main/src/backend/index.ts
```diff
@@ -4,42 +4,71 @@ import { scrapeFinancialAccountsAndFetchTransactions } from '@/backend/import/im
...
+
+export function stopPeriodicScraping() {
+  if (intervalId) {
+    clearInterval(intervalId);
+  }
+}
 
 export async function scrapeAndUpdateOutputVendors(config: Config, optionalEventPublisher?: Events.EventPublisher) {
   const eventPublisher = optionalEventPublisher ?? new Events.BudgetTrackingEventEmitter();
+  await eventPublisher.emit(EventNames.LOG, {
```
brafdlog, line 43: @baruchiro this is a really strange one. It seems that the first event published gets lost. I spent an hour trying to understand why the IMPORT_PROCESS_START doesn't do anything until I added an event before it, and only then it started to work. Couldn't figure out why

packages/main/src/handlers/index.ts
```diff
@@ -67,10 +68,11 @@ export const registerHandlers = () => {
   ipcMain.on('scrape', async (event: IpcMainEvent) => {
     const config = await getConfig();
     const eventSubscriber = new BudgetTrackingEventEmitter();
-    scrapeAndUpdateOutputVendors(config, eventSubscriber);
```
brafdlog, line 70: @baruchiro  we had a bug here, listening to events was set up only after running scrape, and scraping was without await, so there was a race condition on whether or not the first event set would be caught by the onAny handler or not.

baruchiro (approved):

PR #618, opened 2024-11-10
add ascending because of the rate limit

baruchiro:
See:

https://github.com/actions/stale?tab=readme-ov-file#ascending
https://github.com/actions/stale?tab=readme-ov-file#operations-per-run


baruchiro:
@shaiu Please note

PR #619, opened 2024-11-10
New: Add the Otp Modal

shaiu:


packages/preload/src/eventsBridge.ts
```diff
@@ -91,3 +91,7 @@ export async function electronGoogleOAuth2Connector(): Promise<Credentials> {
 export async function createSpreadsheet(spreadsheetId: string, credentials: Credentials): Promise<string> {
   return electron.ipcRenderer.invoke('createSpreadsheet', spreadsheetId, credentials);
 }
+
+export async function sendUserInput(input: string) {
```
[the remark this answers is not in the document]
shaiu: @baruchiro all other methods don't have error handling at all, so I don't think we should start now.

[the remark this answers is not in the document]
baruchiro: הקופיילוט די גרוע, תתעלם

baruchiro (changes requested):
Please fix the issues and I will review it again.

packages/preload/src/eventsBridge.ts
```diff
@@ -91,3 +91,7 @@ export async function electronGoogleOAuth2Connector(): Promise<Credentials> {
...
+  await electron.ipcRenderer.send('get-otp-response', input);
+}
```
baruchiro, line 97:
```suggestion
export async function sendOTPResponse(input: string) {
  await electron.ipcRenderer.send('get-otp-response', input);
}
```

packages/preload/src/eventsBridge.ts
```diff
@@ -91,3 +91,7 @@ export async function electronGoogleOAuth2Connector(): Promise<Credentials> {
 export async function createSpreadsheet(spreadsheetId: string, credentials: Credentials): Promise<string> {
   return electron.ipcRenderer.invoke('createSpreadsheet', spreadsheetId, credentials);
 }
+
+export async function sendUserInput(input: string) {
+  await electron.ipcRenderer.send('get-otp-response', input);
```
baruchiro, line 96: Why `send` and not `invoke`?

packages/renderer/src/components/GetOtp.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { useConfigStore } from '/@/store/ConfigStore';
+import { sendUserInput } from '#preload';
+import { ModalStatus } from '/@/types';
+
+const GetOtp = () => {
+  const configStore = useConfigStore();
+  const [modalStatus, setModalStatus] = useState<ModalStatus>(ModalStatus.HIDDEN);
+  const [inputText, setInputText] = useState('');
+
+  const closeModal = () => setModalStatus(ModalStatus.HIDDEN);
```
baruchiro, line 14: The `ModalStatus` is to reuse the main modal (`Body.tsx`) for different usages.

Here you create a new Modal, I think you can just use a boolean state to decide if the Modal is open or not.

packages/preload/src/eventsBridge.ts
shaiu: I'm still learning. What's the difference?

packages/renderer/src/components/GetOtp.tsx
shaiu: Done. changed to boolean

packages/preload/src/eventsBridge.ts
baruchiro: I'm not sure there is a difference,

https://stackoverflow.com/a/59889863/839513

But all other calls are using `invoke`, so you should as well.

packages/renderer/src/components/GetOtp.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { observer } from 'mobx-react-lite';
+import { useConfigStore } from '/@/store/ConfigStore';
+import { sendOTPResponse } from '#preload';
+
+const GetOtp = () => {
+  const configStore = useConfigStore();
+  const [modalStatus, setModalStatus] = useState<boolean>(false);
+  const [inputText, setInputText] = useState('');
+
+  const closeModal = () => setModalStatus(false);
```
baruchiro, line 13: OK, you allow the user to close the modal, but the backend process will still wait for the OTP.

If the user closed the modal, I assume you should `sendOTPResponse(undefined)` or something to fail the process.

packages/renderer/src/components/GetOtp.module.css
baruchiro, line 1: I think all the styles in this document are not used, and the only used style `modalHeader` is not declared here.

packages/preload/src/eventsBridge.ts
shaiu: Done

packages/renderer/src/components/GetOtp.tsx
shaiu: Done

packages/renderer/src/components/GetOtp.module.css
shaiu: Right. Removed css file

baruchiro (approved):

PR #641, opened 2025-02-08
Build: fix linux build

whatuserever:
Fixes #640: Linux build was failing due to icon. It doesn't seem to work with `ics` or `ico` formats, or at least not the ones which are used for Mac/Windows.
To mitigate, set a png icon.

Note that for some reason the icon does not seem to actually apply:
image
I thought it might be due to the png dimensions, but I tried setting a 512x512 png instead and the result was the same. 

## Summary by CodeRabbit

- **New Features**
	- Updated the Linux application build with a fresh icon for enhanced visual branding.
- **Improvements**
	- Expanded the testing matrix to include Windows as an operating system for running tests.

PR #643, opened 2025-02-09
Fix: Ubuntu GitHub test action

whatuserever:
The `ubuntu-latest` tag was changed to point to Ubuntu 24.04 and has caused regressions. See https://github.com/microsoft/playwright/issues/34251 and https://github.com/electron/electron/issues/42510 for more details.
TL;DR: 
> The solution is to lift the restrictions that Ubuntu 24.04 implements in the AppImages.
> ```
> sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
> ```

## Update
The command above fixed the `Process failed to launch` error, but instead produced a `Timeout 30000ms exceeded` error. For now, setting tests to run on ubuntu-22.04 as that seems to work at the moment.

## Summary by CodeRabbit

- **Chores**
	- Enhanced the automated testing workflow by updating the operating system to Ubuntu 22.04, improving the reliability of our testing process.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,6 +24,13 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4
+      - name: Setup Ubuntu Binary Installation # TODO: Remove when https://github.com/electron/electron/issues/42510 is fixed
+        if: ${{ runner.os == 'Linux' }}
+        run: |
+          if grep -q "Ubuntu 24" /etc/os-release; then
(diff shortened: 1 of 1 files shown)


baruchiro:
@whatuserever thank you!!

It is still not working.

1. It is OK for me to unite both PRs (this and #641) since they are both fixing the Linux pipelines.
2. If there is a bug on Electron that supports Ubuntu 24.04, it is OK for me to test it on the previous Ubuntu version.

whatuserever:
Hmm I'm not sure it's worth merging this PR if it's not solving the issue 🤔

baruchiro:
I think you didn't understand me. Can you contact me on Discord?

PR #646, opened 2025-02-10
New: Skip YNAB export for unmapped accounts

### LINKED ISSUE #644: Feature Request: Don't fail YNAB exporter on unmapped accounts
**Is your feature request related to a problem? Please describe.**
I want to use Caspion to export transactions into YNAB, but I have multiple accounts on the same financial providers and I don't want to export all of them.
This is impossible at the moment. If your YNAB config has unmapped accounts, the YNAB exporter hangs and logs the following:
> Unhandled account number `<accountNumber>`

**Describe the solution you'd like**
Simply ignore transactions from accounts that have no YNAB mapping. Emit a message in the logger indicating which accounts were skipped.

**Describe alternatives you've considered**
I see 2 options:
1. Always ignore unmapped accounts
2. Add an option in the config to allow ignoring unmapped accounts. If it's enabled - skip. Otherwise - error.

**Additional context**
The root cause is the throw statement here:
https://github.com/brafdlog/caspion/blob/3e7c0070c382a4a3578af965e79b898f77f234c0/packages/main/src/backend/export/outputVendors/ynab/ynab.ts#L136-L142

I think the proper solution would be to refactor the YNAB exporter such that it exports transactions by account. This way the skip can be done before processing the transactions, and we'll have a single log line per skipped account instead of a line per transaction from skipped accounts.

Something along the lines of:

```typescript
const createTransactions: ExportTransactionsFunction = async ({ transactionsToCreate, startDate }, eventPublisher) => {
  // ...
  const transactionsByAccount = // mapping from accountNumber to list of transactions from that account
  const results = transactionsByAccount.forEach(createTransactionsByAccount(accountNumber, accountTransactions))
  return sum(results)
}

function createTransactionsByAccount(accountNumber, accountTransactions) {
  if (unmapped(accountNumber)) {
    emit("skipping message") // or throw, depending on the decided approach
    return
  }
  // process transactions as before
}
```

I'm willing to work on this, but let me know what you think first :)

whatuserever:
1. Refactored YNAB exporter to run per account
2. If account is not mapped, skip it instead of throwing

Closes #644. Please read the issue for more details.

## Summary by CodeRabbit

- **New Features**
  - Added account-specific transaction export capabilities, allowing users to filter and export transactions by account number.
  - Enhanced transaction processing with integrated vendor configuration and improved error notifications to better inform users when accounts are unmapped.

### DIFF
--- a/packages/main/src/backend/export/outputVendors/ynab/ynab.ts
+++ b/packages/main/src/backend/export/outputVendors/ynab/ynab.ts
@@ -2,6 +2,7 @@ import {
   OutputVendorName,
   type Config,
   type EnrichedTransaction,
+  type ExportTransactionsForAccountFunction,
   type ExportTransactionsFunction,
   type OutputVendor,
   type YnabAccountDetails,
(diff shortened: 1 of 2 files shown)


packages/main/src/backend/export/outputVendors/ynab/ynab.ts
```diff
@@ -37,14 +38,50 @@ async function initFromToken(accessToken?: string) {
...
   }
-  if (!categoriesMap.size) {
-    await initCategories();
+  try {
+    getYnabAccountIdByAccountNumberFromTransaction(accountNumber);
+  } catch (e) {
+    await emitProgressEvent(eventPublisher, allTransactions, `Account ${accountNumber} is unmapped. Skipping.`);
+    return {
+      exportedTransactionsNum: 0,
+    };
```
whatuserever, line 80: As mentioned in the linked issue, let me know if you'd prefer to have an option in the config for enabling the skipping logic.

packages/main/src/backend/export/outputVendors/ynab/ynab.ts
```diff
@@ -37,14 +38,50 @@ async function initFromToken(accessToken?: string) {
...
   }
-  const transactionsFromFinancialAccount = transactionsToCreate.map(convertTransactionToYnabFormat);
+  const transactionsFromFinancialAccount = allTransactions
+    .filter((v) => v.accountNumber === accountNumber)
```
whatuserever, line 83: I decided to do the filter here instead of in the outer function (`createTransactions`) for 2 reasons:
1. The `emitProgressEvent` function (used below) expects to receive `allTransactions`. I'm not sure how important that is, so I didn't want to change that.
2. This way there's no dependency on the caller.

baruchiro:
sorry it is in my todolist

PR #647, opened 2025-02-10
Breaking: Add support for multiple YNAB budgets

### LINKED ISSUE #645: Feature Request: Support multiple YNAB budgets
**Is your feature request related to a problem? Please describe.**
I'm using YNAB with multiple budgets, and I'm unable to choose which account should be mapped to which budget. They all have to be mapped to the same budget.

**Describe the solution you'd like**
In the YNAB config, change the layout such that the budget is defined per account:
| Account number | Ynab budget id | Ynab account id|
|---|---|---|
|`<accountNumber>`|`<ynabBudgetId>`|`<ynabAccountId>`|

**Describe alternatives you've considered**
None

**Additional context**
I'm willing to work on this :)

whatuserever:
This PR makes the following changes to support multiple YNAB accounts:
1. Adds a `Ynab budget id` column to the config
1. Changes `YnabConfig` to also store `ynabBudgetId` alongside `ynabAccountId`
1. Filters the accounts on the `Ynab account id` column based on the budget selected on that row
1. Uses the "old" budget id as the default value for `Ynab budget id` on new rows
1. Changes logic in `ynab.ts` to be based on the relevant `Ynab budget id`

> [!Important]
> This change will not work with existing configs!
> I'm not sure how to support backward compatibility here due to the `YnabConfig` changes.
> If you have any suggestions do let me know.

Note that this PR uses #646 as a base.
I suggest going through the commits one-by-one when reviewing :)

Closes #645 

### Screenshot
image

## Summary by CodeRabbit

- **New Features**
  - Enhanced transaction export functionality that now filters and processes transactions on a per-account basis with dynamic budget category initialization.
  - Updated account mapping interface, which now includes budget selection for improved financial account integration.
  - Added a configuration option allowing users to enable a React-based UI for a refined experience.

These changes streamline transaction management and upgrade the account mapping workflow for better usability and integration.


packages/main/src/backend/export/outputVendors/ynab/ynab.ts
```diff
@@ -138,48 +174,62 @@ function getYnabAccountIdByAccountNumberFromTransaction(transactionAccountNumber
...
+          name: category.name,
+          category_group_id: category.category_group_id,
+        }))
+        .forEach((category) => {
+          categoriesMap.set(category.name, category);
+        });
+    });
+    budgetCategoriesMap.set(budgetId, categoriesMap);
   });
 }
```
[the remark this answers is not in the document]
whatuserever: Good catch! I'll fix this later today

packages/renderer/src/components/exporters/YnabAccountMappingTable.tsx
```diff
@@ -44,12 +47,29 @@ const YnabAccountMappingTable = ({
...
+    },
     {
       dataField: 'ynabAccountId',
       text: 'Ynab account id',
       editor: {
         type: Type.SELECT,
-        getOptions: () => {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        getOptions: (_: any, { row }: { row: AccountMapping }) => {
+          const budgetId = (row as AccountMapping).ynabBudgetId;
```
whatuserever, line 72: I have no idea how to properly type this.
`getOptions` [isn't covered by DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-bootstrap-table-next/index.d.ts), the [documentation doesn't indicate the type,](https://react-bootstrap-table.github.io/react-bootstrap-table2/docs/basic-celledit.html#editorgetoptions) and [the source code is untyped JavaScript](https://github.com/react-bootstrap-table/react-bootstrap-table2/blob/ef22dc47ebeb985b28980001f806eeb01a687973/packages/react-bootstrap-table2-filter/src/components/select.js#L94-L96). The only hint was [the example here](https://github.com/react-bootstrap-table/react-bootstrap-table2/blob/master/packages/react-bootstrap-table2-editor/README.md#editorgetoptions).

packages/renderer/src/types.tsx
```diff
@@ -70,7 +70,7 @@ export interface GoogleSheetsConfig extends OutputVendorConfigBase {
 export interface YnabConfig extends OutputVendorConfigBase {
   options: {
     accessToken: string;
-    accountNumbersToYnabAccountIds: Record<string, string>;
+    accountNumbersToYnabAccountIds: Record<string, { ynabBudgetId: string; ynabAccountId: string }>;
     budgetId: string;
```
whatuserever, line 74: The reason I'm keeping the "original" `budgetId` is to pre-populate new rows in the config with it. This way users do not have to reselect a budget for every account if they just use a single budget. This `budgetId` is not used at all other than that.
However, I think it might be a bit confusing for the user that they have to define a "main" `budget Id` and an individual `Ynab budget id` per account.

I can see 2 alternative:
1. Rename the title in the UI from `budget id` to `default budget id`, maybe even add a tooltip that explains it's only used for pre-populating new rows.
1. Remove it, and use the budget id of the last row (if exists) for pre-populating new rows.
    * This might be troublesome, because in my experience, the rows do not retain their original order on changes.

packages/renderer/src/components/exporters/YnabAccountMappingTable.tsx
baruchiro: I passed the same way you went through (if I was reading your comment completely before trying to solve this... 😂), and then I found out we don't need to research this.

You can read about this here:
https://stackoverflow.com/a/63254304/839513

We use `any` to **ignore** type checking, which may hide errors, and we try to avoid it.

But if you use `unknown`, it will prevent you from accessing and manipulating the object, which is exactly the situation here, we don't know the type of `_` and we don't want to use it.

packages/renderer/src/types.tsx
```diff
@@ -70,7 +70,7 @@ export interface GoogleSheetsConfig extends OutputVendorConfigBase {
 export interface YnabConfig extends OutputVendorConfigBase {
   options: {
     accessToken: string;
-    accountNumbersToYnabAccountIds: Record<string, string>;
+    accountNumbersToYnabAccountIds: Record<string, { ynabBudgetId: string; ynabAccountId: string }>;
```
baruchiro, line 73: Maybe instead of declaring a `Record`, you can declare an object:

```typescript
accountNumbersToYnabAccountIds: {
    [budgetId: string]: {
        ynabBudgetId: string;
        ynabAccountId: string
    }
}
```

I'm using this structure when it is not clear what is the key, and this give me the option to give a name to the key.

packages/renderer/src/components/exporters/YnabAccountMappingTable.tsx
whatuserever: Sounds good, I'll change this tomorrow

packages/renderer/src/types.tsx
whatuserever: Sure, will do.

btw, the key is `accountNumber`, not `budgetId`. Just further emphasizing your point :sweat_smile:

whatuserever: The linter does not allow this :/

```
error  A record is preferred over an index signature  @typescript-eslint/consistent-indexed-object-style
```

whatuserever: I can do `type AccountNumber = string` and `Record<AccountNumber, ...>` if you find that better

baruchiro:
I need to review this

PR #648, opened 2025-02-15
New: Add config migrations mechanism

whatuserever:
#647 makes a breaking change to the config. To avoid breaking installations with the old config, @baruchiro suggested implementing a migration mechanism.

~~This PR implements it using Zod:~~
* ~~Defined Zod schema for each config version. This avoids creating a dependency on the types used in the "live" code.~~ Used `any` for the config types.
* Added migration logic that iteratively runs migrations starting with the migration relevant to the current config version until there are no more migrations left.
  * Migration functions receive a config of version `n` and return a config of version `n+1`
* Added `version` key to the `Config` type.

## Summary by CodeRabbit

- **New Features**
  - Added configuration versioning support with a new version property for managing updates.
  - Introduced an automatic migration process that seamlessly updates older configuration formats to the current version.
  - Updated default settings and test configurations to integrate the new versioning mechanism.


packages/main/src/backend/configManager/configMigration/configMigrator.ts
```diff
@@ -0,0 +1,31 @@
+import { z } from 'zod';
+import { type Config } from '../../commonTypes';
+import { isOriginalConfig } from './versions/original';
+import { migrateOriginalToV1, v1ConfigSchema } from './versions/v1';
+
+const latestConfigSchema = v1ConfigSchema;
+
+// migrations[n] should be a function that converts version n to version n+1
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const migrations: Record<number, (config: any) => any> = {};
```
whatuserever, line 10: Since `migrations` is gonna hold all migration functions, and each migration function has a different signature (`v1 => v2`, `v2 => v3`, etc...), I'm not sure if it's possible to type this correctly :/

I tried experiment with [`z.discriminatedUnion('version', [...])`](https://zod.dev/?id=discriminated-unions), or defining a shared interface `interface VersionedConfig {version: number}`, but I couldn't get anything to work. If you can think of a way to avoid `any` here please let me know :slightly_smiling_face:

packages/main/src/backend/configManager/configMigration/versions/original.ts
```diff
@@ -0,0 +1,94 @@
...
+  'behatsdaa',
+  'beyahadBishvilha',
+  'oneZero',
+  'pagi',
+]);
+
+export const googleSheetsConfigSchema = z.object({
+  active: z.boolean(),
+  options: z.object({
+    credentials: z.any(),
```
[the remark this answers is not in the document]
whatuserever: `credentials` and `loginFields` are coming from 3rd parties (Google, israeli-bank-scrapers). I didn't want to create a schema for them as they're outside of our control.

packages/main/src/backend/configManager/configMigration/versions/v1.ts
```diff
@@ -0,0 +1,11 @@
+import { z } from 'zod';
+import { originalConfigSchema } from './original';
+
+export const v1ConfigSchema = originalConfigSchema.extend({ version: z.literal(1) });
+
+export function migrateOriginalToV1(v1Config: z.infer<typeof originalConfigSchema>): z.infer<typeof v1ConfigSchema> {
+  return {
+    ...v1Config,
+    version: 1,
+  };
```
baruchiro, line 10: Exactly, this is the migration:

```
const migrated = {
  ...originalConfig,
  version: 1
}
```

You can do it without all the previous `zod` preparations.

whatuserever: Alright, removed all the `zod` stuff and used a simple `any` for all migration code

packages/main/src/backend/configManager/configMigration/versions/v1.ts
```diff
@@ -0,0 +1,9 @@
+// V1 adds the version field to the config object
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function migrateV0ToV1(v0Config: any): any {
+  return {
+    ...v0Config,
+    version: 1,
+  };
+}
```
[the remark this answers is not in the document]
whatuserever: Eh. As @baruchiro pointed out, these migrations are written once and shouldn't be modified again. This is why we're ok with it being untyped (`any`). By that same logic, I don't see the need for tests.

PR #728, opened 2026-02-07
Feat: Add toggle switch to enable/disable accounts from main screen

brafdlog:
## Summary
- Adds a hover-visible toggle switch to each importer and exporter card on the main screen
- Users can quickly include/exclude accounts from scraping without opening the settings dialog
- Toggle is hidden during scraping, matching the behavior of other action buttons
- Widened account cards (200px → 260px) and container (260px → 300px) to accommodate the new toggle

Closes #713

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/renderer/src/components/accounts/Account.tsx
+++ b/packages/renderer/src/components/accounts/Account.tsx
@@ -1,5 +1,6 @@
 import { useMemo } from 'react';
 import Badge from 'react-bootstrap/Badge';
+import Form from 'react-bootstrap/Form';
 import piggyBank from '../../assets/piggy-bank.svg';
 import { type Account as AccountType, type ExporterEndEvent } from '../../types';
 import styles from './Account.module.css';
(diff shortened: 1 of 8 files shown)


packages/renderer/src/components/accounts/Account.tsx
```diff
@@ -48,6 +50,16 @@ export default function Account({ account, actionButtons }: AccountProps) {
...
         />
       ))}
+      {onToggleActive && (
+        <Form.Check
+          className={styles.activeToggle}
+          type="switch"
+          checked={account.active}
+          onChange={onToggleActive}
+          title={account.active ? 'פעיל' : 'לא פעיל'}
+          aria-label={account.active ? 'פעיל' : 'לא פעיל'}
```
baruchiro, line 62: @copilot I see the onToggleActive passed as prop with `!isScraping ? () => configStore.toggleImporterActive(account.id) : undefined`.

This is strange, Maybe we should add a prop for enable/disable the toggle?

dadarara:
> @dadarara feel free to try this branch and tell me if it works for you.

there is no window exe artifact.  I am not very experienced in git. so dont know how to build and have the exe generated. :)

packages/renderer/src/components/accounts/Account.tsx
brafdlog: Fixed

brafdlog:
@dadarara merged, FYI

dadarara:
great work. thank you !! works very well.

